PR #63, opened 2022-01-06
Update license header year to be 2022

NSeydoux:


matthieubosquet (approved):
Same comment as for the VC library, e2e folder wasn’t hit, neither were the config files.

Different from the vc library: the .eslintrc.js has a license header, but the multiline comment style seems wrong (I can’t imagine there’s a good justification to it having double slash comment, but I might miss something).

PR #84, opened 2022-01-21
Prevent dependabot from running CD

NSeydoux:
Dependabot doesn't have access to repository secrets, and as such it cannot run the CD. The current setup had a bug which resulted in workflows failing rather than being skipped.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -14,13 +14,13 @@ env:
 jobs:
   prepare-deployment:
     runs-on: ubuntu-20.04
+    if: github.actor != 'dependabot[bot]'
     outputs:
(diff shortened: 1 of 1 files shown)


ThisIsMissEm (approved):
looks good to me! nice catch!

Might be an idea to share this in our engineering chat, as I didn't realise this could cause issues.

NSeydoux:
Sure, will do. It's something I want to experiment with though, because I know GH added a dependabot sercrets section. However, even if it were possible, I would rather not have dependabot run the CD, which seems unecessary for typical dependency updates.

matthieubosquet (approved):
Haha, yeah!

PR #92, opened 2022-01-28
Add separate linting and audit workflows

NSeydoux:
This separates auditing and linting from the main CI task. These workflows are still gating, but:
- They fail faster is applicable, 
- They only run in once OS and one Node version
- This makes the reporting clearer.


.github/workflows/audit.yml
```diff
@@ -0,0 +1,33 @@
...
+      - run: npm run licenses:check
+      # Some issues identified by npm audit cannot be fixed by us
+      # (i.e. they require an upstream dependency update),
+      # but using `npm-force-resolutions` we can at least override the upstream
+      # dependency when running an audit.
+      # If this task fails, first attempt to run `npm audit fix` or upgrade the
+      # relevant dependency; when that does not work, you can add the transitive
+      # dependency to the `resolutions` field in package.json according to the
+      # instructions in `npm-force-resolutions`'s README.
+      - run: npx npm-force-resolutions && npm audit --audit-level=moderate
```
ThisIsMissEm, line 27: @inrupt/sdk-developers we could also run this with only `--production` which is the dependencies consumers of our code will get.

.github/workflows/cd-teardown.yml
```diff
@@ -9,21 +9,21 @@ jobs:
...
+        # See https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable
+        run: echo "TAG_SLUG=$(echo "${{ github.event.ref }}" | tr -cd '[:alnum:]-')" >> $GITHUB_ENV
+      - name: Remove npm tag for the deleted branch
+        run: |
+          # Unfortunately GitHub Actions does not currently let us do something like
+          #     if: secrets.NPM_TOKEN != ''
+          # so simply skip the command if the env var is not set:
+          if [ -n $NODE_AUTH_TOKEN ]; then npm dist-tag rm @inrupt/solid-client-access-grants $TAG_SLUG; fi
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
ThisIsMissEm, line 28: Generally it's a good idea to keep the secret name and env variable name aligned, as this isn't actually the `NODE_AUTH_TOKEN` but the `NPM_TOKEN` (node doesn't have auth tokens) — the only reason I mention this is because I've seen more complex setups where this becomes an issue.

ThisIsMissEm (approved):

.github/workflows/publish-website.yml
```diff
@@ -11,43 +11,43 @@ jobs:
...
+          git remote add gh-pages-remote https://x-access-token:$GH_TOKEN@github.com/$GITHUB_REPOSITORY.git
+          git fetch --no-recurse-submodules
+          cd docs
+          git worktree add ./gh-pages gh-pages
+          cd gh-pages
+          git rm -r .
+          cp -r ../dist/. .
+          git add .
+          git commit --message="Deploying to GitHub Pages from $GITHUB_SHA"
+          git push gh-pages-remote gh-pages:gh-pages
```
ThisIsMissEm, line 50: It may be an idea to move the larger script blocks like this to scripts within the workflows folder, such that they're not just inline in yaml.

.github/workflows/audit.yml
NSeydoux: This could be useful if the audit becomes an issue, but if we don't have an audit for dev dependencies, couldn't this compromise the built code (like, in a very worst case scenario) ?

ThisIsMissEm: Yeah, I was more thinking for production dependencies, we'd wanna not use npm-force-resolutions, but for dev dependencies use it.

.github/workflows/cd-teardown.yml
NSeydoux: Actually `NODE_AUTH_TOKEN` is the environment variable used by the `setup-node` GH action for publication top NPM: https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#publish-to-npmjs-and-gpr-with-npm

.github/workflows/publish-website.yml
NSeydoux: Fixed in 44c5daf

PR #97, opened 2022-01-31
Use GH environments in end-to-end node tests

NSeydoux:
This refactors the end-to-end tests to use GH environments to set the environment variables, instead of using a single environment with prefixed environment variables and test-side logic to figure out which should be picked up. Note that this also adds a bit of setup/teardown to the tests: the file for which the access request is granted is created at the beginning of the test suite, and deleted at the end.


e2e/node/.env.example
```diff
@@ -1,24 +0,0 @@
...
-
-E2E_TEST_DEV_NEXT_IDP_URL=https://openid.dev-next.inrupt.com
-E2E_TEST_DEV_NEXT_VC_SERVICE=https://vc.dev-next.inrupt.com
-E2E_TEST_DEV_NEXT_REQUESTOR=https://id.dev-next.inrupt.com/devnextrequest
-E2E_TEST_DEV_NEXT_RESOURCE_OWNER=https://id.dev-next.inrupt.com/devnextrowner
-# Obtain these via https://openid.dev-next.inrupt.com/registration.html
-E2E_TEST_ESS_REQUESTOR_CLIENT_ID=5bc380bb-883f-468d-8275-68c903a75766
-E2E_TEST_ESS_REQUESTOR_CLIENT_SECRET=o9SCi3bDSmIqXyJqs6FE1smVMvZDgthK
-E2E_TEST_ESS_RESOURCE_OWNER_CLIENT_ID=57d036e2-5deb-4f71-b1c4-0b82327e9a29
-E2E_TEST_ESS_RESOURCE_OWNER_CLIENT_SECRET=e0fcdpcDsLLitXZTmJKFHhkuZa61B0sY
```
NSeydoux, line 24: 0_0 these should never have been there. I'll revoke these credentials, generate new ones and update the secrets accordingly.

matthieubosquet: oh dear

.github/workflows/e2e-node.yml
```diff
@@ -0,0 +1,46 @@
...
+    if: ${{ github.actor != 'dependabot[bot]' }}
+    runs-on: ${{ matrix.os }}
+    environment:
+      name: ${{ matrix.environment-name }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        node-version: [16.x]
+        # NSS does not support static client registration, which we rely on for testing.
+        environment-name: ["ESS Dev-Next"]
```
matthieubosquet, line 22: What about "ESS Production"?

matthieubosquet (approved):
LGTM

.github/workflows/e2e-node.yml
NSeydoux: Production still running on 1.1, access grants are not implemented yet (that I'm aware of).

e2e/node/e2e.test.ts
```diff
@@ -7,313 +23,269 @@ import {
...
-    requestorClientId,
-    requestorClientSecret,
-    resourceOwnerClientId,
-    resourceOwnerClientSecret
-  ) => {
-    const oidcIssuer = new URL(`https://${oidcIssuerDisplay}`).href;
-    const vcService = new URL(`https://${vcServiceDisplay}`).href;
-    const requestor = new URL(`https://${requestorDisplay}`).href;
-    const resourceOwner = new URL(`https://${resourceOwnerDisplay}`).href;
+describe(`End-to-end access grant tests for environment [${environment}}]`, () => {
```
ThisIsMissEm, line 39: Isn't this self-documenting via the matrix runner?

Screenshot 2022-01-31 at 17 03 41

We could get rid of all the environment validation type logic and instead rely on the runner to be showing the environment? We could also do a check in getEnvironment that if not `process.env.CI` then print out the environment you're executing on as developer diagnostics.

PR #99, opened 2022-02-01
Preview demo template

NSeydoux:
This introduces the browser-based demo app which will be eventually used in the browser-based tests. The present PR only adds the vanilla web demo template and adjusts the NPM scripts accordingly, the demo app will be updated in a future PR to be usable in browser-based e2e tests.


.github/dependabot.yml
```diff
@@ -22,3 +22,10 @@ updates:
     # Check the npm registry for updates every day (weekdays).
     schedule:
       interval: "daily"
+  # Enable version updates for embedded demo app
+  - package-ecosystem: "npm"
+    # Look for `package.json` and `lock` files in the `e2e/browser/webapp` directory.
+    directory: "/e2e/browser/webapp"
+    # Check the npm registry for updates every day (weekdays).
+    schedule:
+      interval: "daily"
```
matthieubosquet, line 31:
```suggestion
      interval: "weekly"
```

Maybe to reduce the noise as it's just a demo app?

On the other hand, you might wanna run the audit daily.

e2e/browser/src/.gitignore
```diff
@@ -0,0 +1,34 @@
...
+# debug
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# local env files
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
```
matthieubosquet, line 31:
```suggestion
# debug
*.log*

# local files
*.local
```

No need to be more specific here. A file with a `.local` suffix will always be something we want to ignore, same for logs.

e2e/browser/src/components/appContainer/index.jsx
```diff
@@ -0,0 +1,85 @@
+/**
+ * Copyright 2021 Inrupt Inc.
```
matthieubosquet, line 2:
```suggestion
 * Copyright 2022 Inrupt Inc.
```

I'm only gonna comment once, but this is a problem in every file.

It looks like the linting of licenses is not properly setup, I would probably try to re-use the license resource from the root folder and remove the one here altogether.

e2e/browser/src/components/header/index.jsx
```diff
@@ -0,0 +1,95 @@
...
+            >
+              Log Out
+            </LinkButton>
+          </LogoutButton>
+        )}
+
+        {!sessionRequestInProgress && !session.info.isLoggedIn && (
+          <LoginButton
+            authOptions={{ clientName: CONFIG.demoTitle }}
+            oidcIssuer="https://broker.pod.inrupt.com"
```
matthieubosquet, line 80: Could this be a config too?

e2e/browser/src/config.js
```diff
@@ -0,0 +1,10 @@
+/* eslint import/extensions: 0 */
+import packageJson from "./package.json";
+
+export default () => ({
+  libraryRepoUrl: packageJson.repository.url,
+  demoRepoUrl: packageJson.repository.url,
+  copyright: "Copyright 2021 Inrupt, Inc.",
```
matthieubosquet, line 7:
```suggestion
  copyright: "Copyright 2022 Inrupt, Inc.",
```

e2e/browser/src/pages/styles.css
```diff
@@ -0,0 +1,61 @@
...
+      format("svg"),
+    /* Legacy iOS */ url("../public/Raleway/Raleway-ExtraBold.ttf")
+      format("truetype"),
+    /* Safari, Android, iOS */ url("../public/Raleway/Raleway-ExtraBold.woff")
+      format("woff"),
+    /* Modern Browsers */ url("../public/Raleway/Raleway-ExtraBold.woff2")
+      format("woff2"); /* Modern Browsers */
+  font-weight: normal;
+  font-style: normal;
+}
```
matthieubosquet, line 54: I'm not sure why we use those fonts in demo apps.

It feels like seriously overly compilcating things (and also requires to include the extra Raleway license).

I think demo apps should be barebones as much as possible.

Being barebones not only reduces work (and maintenance) efforts on our side, it also would allow anyone checking this app out to focus on the relevant parts of this app: i.e. understanding how the library works and can be used.

matthieubosquet (approved):
I would highly recommend making the app barebones (might be a team discussion I guess).

Otherwise, most relevant comment is about linting license header.

As per PR 100, I'm probably not the best person to review apps, but nothing seems particularly wrong.

.github/dependabot.yml
NSeydoux: I'm not sure it would make a huge difference, as I'm merging dependabot PRs daily anyway.

e2e/browser/src/pages/styles.css
NSeydoux: As discussed, this comes from the demo template, and will be replaced by a more barebones app.

e2e/browser/src/components/header/index.jsx
NSeydoux: 6ae0b67a5addc97bb6551192abeb6dfae002feb1

e2e/browser/src/config.js
NSeydoux: 46a17523c4d2a42b7221567ee13ae2d7be023604

PR #100, opened 2022-02-04
Prepare e2e test app

NSeydoux:
This prepares the e2e test app for playwright tests. The app displays a couple of buttons to log in, create a resource, grant access to it, revoke the grant and delete the resource. This should make playwright tests fairly straightforward, just clicking around for setup, testing displayed values and teardown.

One thing I'm unsure about is the linking from the test app to the local code in `solid-client-access-grants`. I assume setting up the end-to-end tests would require to systematically run `npm link` in CI, does that sound correct @ThisIsMissEm ?

### DIFF
--- a/e2e/browser/src/pages/index.jsx
+++ b/e2e/browser/src/pages/index.jsx
@@ -20,14 +20,125 @@
  */
 
 import { useSession } from "@inrupt/solid-ui-react";
+import { useEffect, useState } from "react";
+import {
+  approveAccessRequest,
+  revokeAccessGrant,
+} from "@inrupt/solid-client-access-grants";
+import {
+  getPodUrlAll,
+  saveFileInContainer,
+  getSourceUrl,
+  deleteFile,
+} from "@inrupt/solid-client";
+
+// This is the content of the file uploaded manually at SHARED_FILE_IRI.
+const SHARED_FILE_CONTENT = "Some content.\n";
 
 export default function Home() {
   const { session } = useSession();
+  const [accessGrant, setAccessGrant] = useState(undefined);
+  const [sharedResourceIri, setSharedResourceIri] = useState(undefined);
+
+  const handleCreate = (e) => {
+    // This prevents the default behaviour of the button, i.e. to resubmit, which reloads the page.
+    e.preventDefault();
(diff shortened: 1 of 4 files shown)


e2e/browser/src/components/header/index.jsx
```diff
@@ -77,7 +77,7 @@ export default function Header() {
         {!sessionRequestInProgress && !session.info.isLoggedIn && (
           <LoginButton
             authOptions={{ clientName: CONFIG.demoTitle }}
-            oidcIssuer="https://broker.pod.inrupt.com"
+            oidcIssuer="https://openid.dev-next.inrupt.com"
```
matthieubosquet, line 80:
```suggestion
            oidcIssuer="https://login.inrupt.com"
```

Maybe use the next prod which should be more stable?

e2e/browser/src/pages/index.jsx
```diff
@@ -20,14 +20,125 @@
...
+    if (typeof sharedResourceIri !== "string") {
+      // If the resource does not exist, do nothing.
+      return;
+    }
+    (async () => {
+      const accessGrant = await approveAccessRequest(
+        session.info.webId,
+        undefined,
+        {
+          requestor: "https://johndoe.webid",
```
matthieubosquet, line 94: Don't we need a "real" webid here?

As in, is there no WebID verification step?

matthieubosquet (approved):
I'm probably not the best to review apps, but there is nothing specific that caught my attention as wrong.

I'm not sure I understand how the end to end test works without valid testing credentials.

e2e/browser/src/components/header/index.jsx
NSeydoux: Currently it appears that editing one's WebID with https://id.inrupt.com/ doesn't work, so I'll keep using dev-next and update this when it is fixed.

e2e/browser/src/pages/index.jsx
NSeydoux: No, there is no verification step when granting access, the verification would happen when trying to exercise this grant. Similarly, you could add any arbitrary IRI in an ACP matcher, there is no verification step at this point in the process.

PR #104, opened 2022-02-09
Setup browser-based end-to-end tests with playwright

NSeydoux:
This adds browser-based tests on top of the test app introduced in previous commits. These tests are driven by playwright, and use a slightly different approach to what we have been doing before: instead of calling to helpers injected from the browser page to the tests runner, this is a more "traditional" UI test, where the test driver simply clicks on buttons and waits for UI changes. This is a bit less flexible than the previous approach, but it is easier to understand.

It has already been agreed that the current test app is not really fit for this purpose, and instead of the full-blown web demo UI we'd rather use a much more bare-bone app. This will be addressed in future commits, and hopefully would have a limited impact on the tests, which only rely on CSS selectors.

EDIT: The tests currently don't run against the local lib, which I think could be done using `npm link` but doesn't work at the moment.


package.json
```diff
@@ -13,7 +13,7 @@
     "test": "jest src",
     "test:e2e": "jest --config=jest.e2e.config.js e2e/node",
     "test:e2e:browser": "playwright test",
-    "test:e2e:browser:build": "npm link && cd e2e/browser/src && npm ci && npm link @inrupt/solid-client-access-grants",
+    "test:e2e:browser:build": "cd e2e/browser/src && npm ci",
```
ajacksified, line 16: I think what you want is `cd e2e/browser/src && npm ci && npm link ../../../`

.github/workflows/e2e-browser.yml
```diff
@@ -0,0 +1,65 @@
...
+        uses: actions/setup-node@v2.5.1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Cache node modules
+        uses: actions/cache@v2.1.7
+        env:
+          cache-name: cache-node-modules
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-node${{ runner.node-version }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
```
matthieubosquet, line 35:
```suggestion
    runs-on: ubuntu-latest
    environment:
      name: ${{ matrix.environment-name }}
    strategy:
      matrix:
        # NSS does not support static client registration, which we rely on for testing.
        environment-name: ["Inrupt Dev-Next", "Inrupt Production"]
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-node@v2
        with:
          node-version: "16"
          cache: "npm"
```

The setup node action already deals with caching: https://github.com/marketplace/actions/setup-node-js-environment#caching-packages-dependencies.

 Let's not use a matrix when we don't need it (I know it's a mistake we have in other CI, but let's not reproduce it here, one less to fixup ^^,).
 
 Let's maybe use a set of standardised environment names across repos: https://github.com/inrupt/solid-client-js/settings/environments.
 
 I think we should test against Inrupt Production and potentially only make that one gating.
 
 There is literally no interest in making GitHub Actions versions so specific. Workflows will not break because of minor/patch changes and it just creates unnecessary dependabot noise.

.github/workflows/e2e-browser.yml
```diff
@@ -0,0 +1,65 @@
...
+          name: playwright-output
+          path: test-results/
+      - name: Archive code coverage results
+        uses: actions/upload-artifact@v2.3.1
+        continue-on-error: true
+        with:
+          name: code-coverage-report
+          path: coverage
+      - name: Archive production artifacts
+        uses: actions/upload-artifact@v2.3.1
```
matthieubosquet, line 61:
```suggestion
        uses: actions/upload-artifact@v2
```

.github/workflows/e2e-browser.yml
```diff
@@ -0,0 +1,65 @@
...
+      - run: npx playwright install-deps
+      - run: npx playwright install
+      - run: npm run test:e2e:browser:build
+      - run: npm run test:e2e:browser
+        env:
+          E2E_TEST_ENVIRONMENT: ${{ matrix.environment-name }}
+          E2E_TEST_UI_LOGIN: ${{ secrets.E2E_TEST_UI_LOGIN }}
+          E2E_TEST_UI_PASSWORD: ${{ secrets.E2E_TEST_UI_PASSWORD }}
+      - name: Archive browser-based end-to-end test request logs
+        uses: actions/upload-artifact@v2.3.1
```
matthieubosquet, line 48:
```suggestion
        uses: actions/upload-artifact@v2
```

.github/workflows/e2e-browser.yml
```diff
@@ -0,0 +1,65 @@
...
+        continue-on-error: true
+        with:
+          name: playwright-output
+          path: test-results/
+      - name: Archive code coverage results
+        uses: actions/upload-artifact@v2.3.1
+        continue-on-error: true
+        with:
+          name: code-coverage-report
+          path: coverage
```
matthieubosquet, line 59:
```suggestion
```

I don't think that code coverage is or should be reported by e2e tests.

playwright.config.ts
```diff
@@ -0,0 +1,66 @@
...
+  retries: 1,
+  globalSetup: require.resolve("./e2e/browser/test/globalSetup.ts"),
+  use: {
+    baseURL: "http://localhost:1234",
+    headless: true,
+    screenshot: "only-on-failure",
+    trace: "on",
+    video: "on-first-retry",
+  },
+  projects: [
```
matthieubosquet, line 35: ^^, interesting that you have to pass in an array of projects to set browsers... I'm wondering if there's another way of passing that in.

matthieubosquet (approved):
A couple of amendments I think to the workflow should be made, I described my suggestions so it's clear, but happy to chat if it helps.

Otherwise, LGTM!

.github/workflows/ci.yml
```diff
@@ -18,13 +18,6 @@ jobs:
         uses: actions/setup-node@v2.5.1
         with:
           node-version: ${{ matrix.node-version }}
```
matthieubosquet, line 20:
```suggestion
          node-version: ${{ matrix.node-version }}
          cache: "npm"
```

Don't forget to specify the cache when removing the unnecessary action! ;)

playwright.config.ts
NSeydoux: This looks like the recommanded approach: https://playwright.dev/docs/test-configuration#multiple-browsers

.github/workflows/ci.yml
NSeydoux: Oops, good catch

PR #107, opened 2022-02-14
Upgrade jest mocked

NSeydoux:
This gets ahead of a deprecation warning.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "test:e2e:browser:build": "cd e2e/browser/src && npm ci",
     "licenses:list": "license-checker --production --csv --out LICENSE_DEPENDENCIES_ALL",
     "licenses:check": "license-checker --production --failOn \"AGPL-1.0-only; AGPL-1.0-or-later; AGPL-3.0-only; AGPL-3.0-or-later; Beerware; CC-BY-NC-1.0; CC-BY-NC-2.0; CC-BY-NC-2.5; CC-BY-NC-3.0; CC-BY-NC-4.0; CC-BY-NC-ND-1.0; CC-BY-NC-ND-2.0; CC-BY-NC-ND-2.5; CC-BY-NC-ND-3.0; CC-BY-NC-ND-4.0; CC-BY-NC-SA-1.0; CC-BY-NC-SA-2.0; CC-BY-NC-SA-2.5; CC-BY-NC-SA-3.0; CC-BY-NC-SA-4.0; CPAL-1.0; EUPL-1.0; EUPL-1.1; EUPL-1.1;  GPL-1.0-only; GPL-1.0-or-later; GPL-2.0-only;  GPL-2.0-or-later; GPL-3.0; GPL-3.0-only; GPL-3.0-or-later; SISSL;  SISSL-1.2; WTFPL\"",
+    "prepare": "husky install",
     "prepublishOnly": "npm run build",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -17,7 +17,6 @@
     "licenses:list": "license-checker --production --csv --out LICENSE_DEPENDENCIES_ALL",
     "licenses:check": "license-checker --production --failOn \"AGPL-1.0-only; AGPL-1.0-or-later; AGPL-3.0-only; AGPL-3.0-or-later; Beerware; CC-BY-NC-1.0; CC-BY-NC-2.0; CC-BY-NC-2.5; CC-BY-NC-3.0; CC-BY-NC-4.0; CC-BY-NC-ND-1.0; CC-BY-NC-ND-2.0; CC-BY-NC-ND-2.5; CC-BY-NC-ND-3.0; CC-BY-NC-ND-4.0; CC-BY-NC-SA-1.0; CC-BY-NC-SA-2.0; CC-BY-NC-SA-2.5; CC-BY-NC-SA-3.0; CC-BY-NC-SA-4.0; CPAL-1.0; EUPL-1.0; EUPL-1.1; EUPL-1.1;  GPL-1.0-only; GPL-1.0-or-later; GPL-2.0-only;  GPL-2.0-or-later; GPL-3.0; GPL-3.0-only; GPL-3.0-or-later; SISSL;  SISSL-1.2; WTFPL\"",
     "prepublishOnly": "npm run build",
-    "prepare": "husky install",
```
ajacksified, line 20: Do you also want to remove husky as a dependency?

NSeydoux: True, actually I'm not sure this specific change is necessary as part of this PR. I'll remove husky in a separate PR.

ThisIsMissEm: We did agree to remove it as a dependency, didn't we?

ThisIsMissEm: (though, now looking closer, I realise this does not do that)

NSeydoux: To be precise this change is due to me misunderstanding where an issue was coming from using `npm link`: it complained about husky, but the problem was in the link target, and not this package

NSeydoux: Fixed this in 88ab35d, and will properly remove husky in a separate PR

matthieubosquet (approved):
LGTM!
But not sure if we need the extra import.

src/verify/isValidAccessGrant.test.ts
```diff
@@ -26,9 +26,7 @@ import {
 // This rule complains about the `@jest/globals` variables overriding global vars:
 // eslint-disable-next-line no-shadow
 import { jest, describe, it, expect } from "@jest/globals";
-// This ESLint plugin seems to not be able to resolve subpackage imports:
-// eslint-disable-next-line import/no-unresolved
-import { mocked } from "ts-jest/utils";
+import { mocked } from "jest-mock";
```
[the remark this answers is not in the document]
NSeydoux: Good catch, the additional export has maybe been added to ease migration from `ts-jest/utils`. The PR doesn't specify if one is preferred over the other: https://github.com/facebook/jest/pull/12089. See 8ecad8a.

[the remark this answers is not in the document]
matthieubosquet: :)
Eventually, we'll get rid of that ugly 
```js
// eslint-disable-next-line no-shadow
import { jest, describe, it, expect } from "@jest/globals";
```
as well ^^,

PR #108, opened 2022-02-15
Deprecate passing the VC subject in the access flow

NSeydoux:
This deprecates passing the VC subject to VC issuing functions part of the access request/grant flow. It aligns with the changes implemented in https://github.com/inrupt/solid-client-vc-js/pull/271, which are soon to be released (the tests have been implemented using `npm link locally`). 

This is a non-breaking change, the deprecated signatures are maintained in the API.

Overall, it feels like the codebase is a bit convoluted, and that it should have been a simpler change than what it actually was, but we can address the tech debt later. 

# Checklist

### DIFF
--- a/src/manage/approveAccessRequest.ts
+++ b/src/manage/approveAccessRequest.ts
@@ -161,16 +161,7 @@ export async function approveAccessRequest(
   requestOverride?: Partial<ApproveAccessRequestOverrides>,
   options?: AccessBaseOptions
 ): Promise<VerifiableCredential>;
-/**
- * @deprecated Please remove the `resourceOwner` parameter.
- */
-export async function approveAccessRequest(
-  resourceOwner: WebId,
-  // If the VC is specified, all the overrides become optional
-  requestVc: VerifiableCredential | URL | UrlString,
-  requestOverride?: Partial<ApproveAccessRequestOverrides>,
-  options?: AccessBaseOptions
-): Promise<VerifiableCredential>;
+
 /**
  * Approve an access request. The content of the approved access request is provided
  * as a set of claims, and no input Verifiable Credential is expected.
@@ -187,8 +178,22 @@ export async function approveAccessRequest(
   requestOverride: ApproveAccessRequestOverrides,
   options?: AccessBaseOptions
 ): Promise<VerifiableCredential>;
+
 /**
  * @deprecated Please remove the `resourceOwner` parameter.
+ * @hidden
+ */
+export async function approveAccessRequest(
+  resourceOwner: WebId,
+  // If the VC is specified, all the overrides become optional
+  requestVc: VerifiableCredential | URL | UrlString,
+  requestOverride?: Partial<ApproveAccessRequestOverrides>,
+  options?: AccessBaseOptions
+): Promise<VerifiableCredential>;
+
+/**
+ * @deprecated Please remove the `resourceOwner` parameter.
+ * @hidden
  */
 export async function approveAccessRequest(
   resourceOwner: WebId,
--- a/src/manage/approveAccessRequest.ts
+++ b/src/manage/approveAccessRequest.ts
@@ -85,43 +85,24 @@ async function addVcMatcher(
   );
 }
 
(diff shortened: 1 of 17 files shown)


ThisIsMissEm:
The docs build seems to be failing due to the destruction workaround in #269 — https://vercel.com/inrupt/solid-client-access-grants-js/EFuvyt3LaPkoTJtGxipTBJCxCJNJ

src/manage/approveAccessRequest.ts
```diff
@@ -161,7 +141,92 @@ export async function approveAccessRequest(
...
-  return issueAccessVc(resourceOwner, grantBody, options);
+  return issueAccessVc(grantBody, options);
+}
+
+/**
+ * Approve an access request. The content of the approved access request is provided
+ * as a Verifiable Credential which properties may be overriden if necessary.
+ *
+ * @param requestVc The Verifiable Credential representing the Access Request. If
+ * not conform to an Access Request, the function will throw.
```
matthieubosquet, line 152:
```suggestion
 * not conforming to an Access Request, the function will throw.
```

src/manage/approveAccessRequest.ts
```diff
@@ -161,7 +141,92 @@ export async function approveAccessRequest(
...
+ * @deprecated Please remove the `resourceOwner` parameter.
+ */
+export async function approveAccessRequest(
+  resourceOwner: WebId,
+  // If the VC is specified, all the overrides become optional
+  requestVc: VerifiableCredential | URL | UrlString,
+  requestOverride?: Partial<ApproveAccessRequestOverrides>,
+  options?: AccessBaseOptions
+): Promise<VerifiableCredential>;
+/**
```
matthieubosquet, line 174: Could we consider removing deprecated functions and releasing a minor?

We're still at 0.x and I don't think it makes sense to try and be forward compatible here. It polutes the code (and only pod browser is really consuming this)... Even not considering the lack of consumers, I think a 0.x library should generally really remain clean and remove obsolete function signatures as soon as they become such.

Is there even a live VC issuer that would accept the result of such function calls?

src/manage/approveAccessRequest.ts
```diff
@@ -161,7 +141,92 @@ export async function approveAccessRequest(
...
+ * @deprecated Please remove the `resourceOwner` parameter.
+ */
+export async function approveAccessRequest(
+  resourceOwner: WebId,
+  requestVc: undefined,
+  // If the VC is undefined, then some of the overrides become mandatory
+  requestOverride: ApproveAccessRequestOverrides,
+  options?: AccessBaseOptions
+): Promise<VerifiableCredential>;
```
matthieubosquet, line 199: Same, I think we should remove this altogether.

matthieubosquet (approved):
LGTM

But I think we should remove obsolete function signatures, not deprecate them.

src/manage/approveAccessRequest.ts
ThisIsMissEm: I think I'd be inclined to agree that we can just deprecate and bump minor: https://jubianchi.github.io/semver-check/#/^0.4.1/0.5.0 (though the text there on the constraint is confusing) — pod browser depends on `^0.4.1` which won't accidentally upgrade to `0.5.0` unlike say `^1.2.0` matching up to, but not including, `2.0.0-0`

ThisIsMissEm: (I was pretty confused reading this code earlier and wasn't sure how the docs would end up with this change)

NSeydoux: Initially, the VC subject was included in the request but ignored by the issuer. Now, the deprecated signature maps to the correct function call ignoring the deprecated parameter, so the VC issuer wouldn't see any difference.

NSeydoux: I'll double check how the docs look indeed, I was wondering about that too.

NSeydoux:
Annoyingly, I can't reproduce the docs bug locally.

PR #111, opened 2022-02-17
Redirect to requestor

NSeydoux:
This function abstracts away the query parameter returned to the requestor via the redirect. It is very similar to `redirectToAccessManagementUi`, which is already part of the API, except it's meant for the access management app to call in order to redirect users to the client requesting access with the grant VC IRI.

Note that returning the VC IRI, and not the VC value, is not consistent with the current behaviour, but it is the behaviour we actually desire. The redirection from the requestor to the access management will be fixed in a different PR, and functions abstracting the interaction with query params will be added too.

Open question for reviewers: should redirecting always return a never resolving promise, or should it do so only in the browser ? In this case, it would make sense that the redirect callback provided in Node would be async, and redirecting would return the result of that callback. 

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,18 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 The following changes have been implemented but not released yet:
 
+### New features
+
+- `redirectToRequestor` has been added to the `./manage` module to help access management
+  apps redirect their users to the clients which requested access to a Resource.
+
 ### Bugfixes
 
 - The JSON-LD context URL has changed from https://vc.inrupt.com/credentials/v1 to https://vc.inrupt.com/credentials/v1.
+- `redirectToAccessManagementUi` now returns a promise that never resolves. This
+  prevents from `await`ing on it, have it resolve, and start running code in the underterministic
+  time between the moment when `window.location.href` is set (or the redirect callback
+  is called) and the moment the redirection actually happens.
 
 ### Deprecation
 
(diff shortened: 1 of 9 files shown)


matthieubosquet (approved):
> Open question for reviewers: should redirecting always return a never resolving promise, or should it do so only in the browser?

I would generally avoid never resolving promises. But I don't know enough about how the reasons for it to happen in that way in the browser to express a very educated opinion.

I'm actually wondering whether those functions actually make sense in the node server context?

NSeydoux:
> > Open question for reviewers: should redirecting always return a never resolving promise, or should it do so only in the browser?

> I would generally avoid never resolving promises. But I don't know enough about how the reasons for it to happen in that way in the browser to express a very educated opinion.

In this case, it makes sure people don't rely on something to happen based on a race condition. Setting `window.location` doesn't redirect instantly, some client-side code may still be executed before redirection. Which means if the promise resolves, it would be possible to write 
```
await redirectToAccessManagementUi(...)
doSomething();
```

and face bugs where `doSomething` is sometimes called, sometimes not. `redirectToAccessManagementUi` never resolving enforces that no code is executed beyond this point, as intended by the redirection.

> I'm actually wondering whether those functions actually make sense in the node server context?

I think they do: essentially, both these functions are just an abstraction layer on top of a redirection with a couple of query parameters. If it is relevant to provide this abstraction in the browser, I would argue it should also be provided in Node. The access requestor/access management app may be client-side or server-side apps, the only difference from our point of view being the redirection mechanism, which is native to the browser and framework-specific in Node.

PR #113, opened 2022-02-21
Query param accessors

NSeydoux:
This adds a couple of query parameters accessors to abstract away their actual values to users. Namely, instead of looking at a given query parameter names, users should call the accessor with the redirect IRI as an input. The accessors are async, because the VCs for access request and access grant are expected to be exchanged as IRIs, and the accessors dereference these in order to return the VC value.

Note that these changes are backwards-compatible: for explicitness, the query params names are different if the VC is passed as a value (e.g. `accessGrant`) or as an IRI (e.g. `accessGrantUrl`). The accessors currently work with both, and removing support for the legacy approach can be done in a future major release.

# Checklist

- [X] Relevant documentation, if any, has been written/updated.
- [X] New functions/types have been exported in `index.ts`, if applicable.

### DIFF
--- a/src/manage/getRequestFromRedirectUrl.ts
+++ b/src/manage/getRequestFromRedirectUrl.ts
@@ -60,7 +60,7 @@ export async function getRequestFromRedirectUrl(
     );
   }
 
-  // Legacy: Get the Access Request. The IRI should be used instead, which is why
+  // DEPRECATED: Get the Access Request. The IRI should be used instead, which is why
   // this parameter missing doesn't result in an exception.
   const accessRequestValue = redirectUrlObj.searchParams.get(
     REQUEST_VC_PARAM_NAME
--- a/src/manage/getRequestFromRedirectUrl.ts
+++ b/src/manage/getRequestFromRedirectUrl.ts
@@ -1,0 +1,97 @@
+/**
+ * Copyright 2022 Inrupt Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to use,
+ * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
+ * Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+ * OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+import { UrlString } from "@inrupt/solid-client";
+import {
+  isVerifiableCredential,
+  getVerifiableCredential,
+  VerifiableCredential,
+} from "@inrupt/solid-client-vc";
+import {
+  REDIRECT_URL_PARAM_NAME,
+  REQUEST_VC_PARAM_NAME,
+  REQUEST_VC_URL_PARAM_NAME,
+} from "../discover/redirectToAccessManagementUi";
+import { getSessionFetch } from "../util/getSessionFetch";
+
+/**
+ * Get the Access Grant out of the incoming redirect from the Access Management app.
+ *
+ * @param redirectUrl The URL the user has been redirected to from the access
+ * management app.
+ * @param option Optional properties to customise the behaviour:
+ * - fetch: an authenticated fetch function. If not provided, the default session
+ * from @inrupt/solid-client-authn-browser will be used if available.
+ * @returns An Access Grant
+ */
+export async function getRequestFromRedirectUrl(
+  redirectUrl: UrlString,
+  option: { fetch?: typeof fetch } = {}
+): Promise<{ accessRequest: VerifiableCredential; redirectUrl: UrlString }> {
+  const redirectUrlObj = new URL(redirectUrl);
+  const authFetch = option.fetch ?? (await getSessionFetch(option));
+
+  // Get the URL where the requestor expects the user to be redirected with
+  // the grant.
+  const requestorRedirectUrl = redirectUrlObj.searchParams.get(
+    REDIRECT_URL_PARAM_NAME
+  );
+  if (requestorRedirectUrl === null) {
+    throw new Error(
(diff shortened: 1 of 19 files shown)


src/manage/getRequestFromRedirectUrl.ts
```diff
@@ -0,0 +1,106 @@
...
+ * @param redirectUrl The URL the user has been redirected to from the access
+ * management app.
+ * @param option Optional properties to customise the behaviour:
+ * - fetch: an authenticated fetch function. If not provided, the default session
+ * from @inrupt/solid-client-authn-browser will be used if available.
+ * @returns An Access Grant
+ */
+export async function getRequestFromRedirectUrl(
+  redirectUrl: UrlString,
+  option: { fetch?: typeof fetch } = {}
```
matthieubosquet, line 46:
```suggestion
  options: { fetch?: typeof fetch } = {}
```

src/request/getGrantFromRedirectUrl.ts
```diff
@@ -0,0 +1,83 @@
...
+ * @param redirectUrl The URL the user has been redirected to from the access
+ * management app.
+ * @param option Optional properties to customise the behaviour:
+ * - fetch: an authenticated fetch function. If not provided, the default session
+ * from @inrupt/solid-client-authn-browser will be used if available.
+ * @returns An Access Grant
+ */
+export async function getGrantFromRedirectUrl(
+  redirectUrl: UrlString,
+  option: { fetch?: typeof fetch } = {}
```
matthieubosquet, line 45:
```suggestion
  options: { fetch?: typeof fetch } = {}
```

src/manage/getRequestFromRedirectUrl.ts
```diff
@@ -0,0 +1,106 @@
...
+} from "@inrupt/solid-client-vc";
+import {
+  REDIRECT_URL_PARAM_NAME,
+  REQUEST_VC_PARAM_NAME,
+  REQUEST_VC_URL_PARAM_NAME,
+} from "../discover/redirectToAccessManagementUi";
+import { getSessionFetch } from "../util/getSessionFetch";
+
+/**
+ * Get the Access Grant out of the incoming redirect from the Access Management app.
```
matthieubosquet, line 35:
```suggestion
 * Get the Access Request out of the incoming redirect from the Access Management app.
```

src/manage/getRequestFromRedirectUrl.ts
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * @param redirectUrl The URL the user has been redirected to from the access
+ * management app.
+ * @param option Optional properties to customise the behaviour:
+ * - fetch: an authenticated fetch function. If not provided, the default session
+ * from @inrupt/solid-client-authn-browser will be used if available.
+ * @returns An Access Grant
+ */
+export async function getRequestFromRedirectUrl(
```
matthieubosquet, line 44:
```suggestion
export async function getAccessRequestFromRedirectUrl(
```

src/request/getGrantFromRedirectUrl.ts
```diff
@@ -0,0 +1,83 @@
...
+ * Get the Access Grant out of the incoming redirect from the Access Management app.
+ *
+ * @param redirectUrl The URL the user has been redirected to from the access
+ * management app.
+ * @param option Optional properties to customise the behaviour:
+ * - fetch: an authenticated fetch function. If not provided, the default session
+ * from @inrupt/solid-client-authn-browser will be used if available.
+ * @returns An Access Grant
+ */
+export async function getGrantFromRedirectUrl(
```
matthieubosquet, line 43:
```suggestion
export async function getAccessGrantFromRedirectUrl(
```

src/request/getGrantFromRedirectUrl.ts
```diff
@@ -0,0 +1,83 @@
...
+            `Parsing the Access Grant [${accessGrantIri}] as JSON failed: ${e.toString()}`
+          );
+        }
+      });
+  if (!isVerifiableCredential(accessGrant)) {
+    throw new Error(
+      `The value received from [${accessGrantIri}] is not a Verifiable Credential`
+    );
+  }
+  return accessGrant;
```
matthieubosquet, line 80: This should be extracted as a `getVerifiableCredential` function call (and used by both getAccessGrant and getAccessRequest).

src/manage/getRequestFromRedirectUrl.ts
```diff
@@ -0,0 +1,106 @@
...
+  redirectUrl: UrlString,
+  option: { fetch?: typeof fetch } = {}
+): Promise<{ accessRequest: VerifiableCredential; redirectUrl: UrlString }> {
```
matthieubosquet, line 47:
```suggestion
): Promise<{ accessRequest: VerifiableCredential; accessRequestorRedirectUrl: UrlString }> {
```

The reason for which the redirectUrl return value is included here is a bit unclear.

It would be good to explain in the comment for the function...

Also, ultimately, I would probably rather provide a function to extract search parameters and standardise the function to retrieve a VC (which could become an extremely thin wrapper around 1 standard getVerifiableCredential function taking in a request param object which can be reused).

matthieubosquet (approved):
I think a lot of code can be put in common between those two functions.

Also, minor comment, the `option` parameter is usually adopting the plural form `options`, we should remain consistent!

Otherwise, it seems to be doing the right thing.

NSeydoux:
Regarding the VC dereferencing, you're right, I wasn't sure to do it as part of this PR but it should be factored out. Actually, I even thing it should be implemented by the VC library, and merely reused here. I'll implement this and update the PR accordingly.

src/manage/getRequestFromRedirectUrl.ts
NSeydoux: This is quite similar to the OIDC flow: the requestor redirects the user to their access management app, and doing so they specify where the access management app should redirect the user back when access has been granted or denied. Agreed on factoring the function to get a VC out.

NSeydoux:
I converted this to draft until https://github.com/inrupt/solid-client-vc-js/pull/284 is reviewed, merged and released.

src/manage/getRequestFromRedirectUrl.ts
```diff
@@ -0,0 +1,106 @@
...
+            `Parsing the Access Grant [${accessRequestIri}] as JSON failed: ${e.toString()}`
+          );
+        }
+      });
+
+  if (!isVerifiableCredential(accessRequest)) {
+    throw new Error(
+      `The value received from [${accessRequestIri}] is not a Verifiable Credential`
+    );
+  }
```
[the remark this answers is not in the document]
NSeydoux: See https://github.com/inrupt/solid-client-vc-js/pull/284

src/request/getGrantFromRedirectUrl.ts
NSeydoux: See https://github.com/inrupt/solid-client-vc-js/pull/284

src/request/getGrantFromRedirectUrl.ts
NSeydoux: e75a4fdc1a5f1fedcc482d6c02c24078794449d8

src/manage/getRequestFromRedirectUrl.ts
NSeydoux: e75a4fdc1a5f1fedcc482d6c02c24078794449d8

src/manage/getRequestFromRedirectUrl.ts
NSeydoux: Renamed in 45b6326446b33707ea2c197f47aedfd4da19cc91

src/manage/getRequestFromRedirectUrl.ts
NSeydoux: 7f6ddc9237f7e268704dfc135703c4546d7ec411

PR #114, opened 2022-02-22
Remove deprecated consent

NSeydoux:
Removing deprecated functions whose name rely on the "consent" terminology, which we are walking away from.


matthieubosquet (approved):
This looks good, but there are a lot of places in the code and comments, including, most importantly in the types (see ConsentApiBaseOptions and consentEndpoint).

Which should also be removed.

PR #115, opened 2022-02-22
Fix release e2e tests

NSeydoux:
The end-to-end tests running after a release are currently failing. This intends at fixing them (but it won't be sure they are actually fixed until the next release)


matthieubosquet (approved):
LGTM

PR #138, opened 2022-03-07
Fix publish-website workflow

ajacksified:
Switch from build-docs-preview-site to preview:docs.


matthieubosquet (approved):
I would probably find the following namespacing more intuitive (`package.json`):

```javascript
"docs:build": "typedoc",
"docs:preview": "npm run build-api-docs; cd docs/api; make html dist",
```

But looks good.

PR #165, opened 2022-03-22
Use default fetch consistently approving request

NSeydoux:
When approving an access request, the library must also change the ACR associated to the resource so that it explicitly supports being accessed by an agent providing an Access Grant. In the case when the calling library is relying on `@inrupt/solid-client-authn-browser` being present and picked up by the VC library, the same assumption is now extended to modifying the ACR: previously, the call to the ACR was unauthenticated. Now, if sca-b is in the app's dependencies, the default session is picked up.


matthieubosquet (approved):
LGTM

ThisIsMissEm (approved):

PR #173, opened 2022-03-22
Uninstall husky and lint-staged

NSeydoux:
Enforcing linting is moved to the IDE and checked for in CI, but husky introduces some undesired complexity in some cases.


ThisIsMissEm (approved):
LGTM, Make sure we have linting and prettier checks on CI

PR #174, opened 2022-03-24
Fixes deserialization of the VC request

NSeydoux:
The VC was deserialized expecting an URL-encoded value, instead of a base-64 encoded one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,8 @@ The following changes have been implemented but not released yet:
 - `approveAccessRequest` wasn't using the default session from `@inrupt/solid-client-authn-browser`
   to set the ACR access appropriately, resulting in `401 Unauthenticated` errors.
 - `getAccessRequestFromRedirectUrl` was deserializing the received VC expecting
-an URL-encoded value, instead of a base64-encoded value, which is how it is serialized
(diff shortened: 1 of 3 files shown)


matthieubosquet (approved):
LGTM

ThisIsMissEm (approved):
LGTM, but is this an interface that is defined in the spec anywhere?

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ The following changes have been implemented but not released yet:
   change is transparent to users.
 - `approveAccessRequest` wasn't using the default session from `@inrupt/solid-client-authn-browser`
   to set the ACR access appropriately, resulting in `401 Unauthenticated` errors.
+- `getAccessRequestFromRedirectUrl` was deserializing the received VC expecting
+an URL-encoded value, instead of a base64-encoded value, which is how it is serialized
+when redirecting the user to the VC management app.
```
ThisIsMissEm, line 25: make sure you run lint & prettier on this file

NSeydoux:
@ThisIsMissEm no, this is purely internal really, this redirection-based flow is not something relying on a defined spec, which is why the query params managements has been absorbed in the API functions: devs should not have to meddle with these parameters manually if we can avoid it.

PR #184, opened 2022-03-28
Prevent a linting failure in CI

NSeydoux:
The example app should be linked to the local version of the code. ESLint complains that it is unresolved. Also, I enforced the linting to be a mandatory check (which I thought it was), to prevent issues like this to get accidentally merged by enabling auto-merge (the issue did not happen locally).


ThisIsMissEm:
I think we might be able to still do `npm install --save @inrupt/solid-client-access-grants` in this example, but then override it with `npm link @inrupt/solid-client-access-grants` — that is, in both cases it'll be present

NSeydoux:
I haven't managed to make ESlint behave and resolve subprojects dependencies form the appropriate `node_modules` folder. I'll leave it for now, not to spend too much time on it when it's not that big of a deal.

ThisIsMissEm:
@NSeydoux shall we just add to our eslintrc `// eslint-disable-next-line import/no-unresolved` as a default to warning rather than error for now?

NSeydoux:
The vercel failure is related to https://github.com/inrupt/solid-client-access-grants-js/pull/187

package.json
```diff
@@ -8,8 +8,8 @@
     "build-api-docs": "typedoc",
     "lint:fix": "npm run lint:eslint -- --fix && npm run lint:prettier -- --write",
     "lint:check": "npm run lint:eslint && npm run lint:prettier -- --check",
-    "lint:eslint": "eslint --config .eslintrc.js \"src/\" \"examples/\"",
-    "lint:prettier": "prettier \"{src,examples}/*.{ts,tsx,js,jsx,css}\" \"**/*.{md,mdx,yml}\"",
+    "lint:eslint": "eslint --config .eslintrc.js \"src/\" \"e2e/\" \"examples/\"",
+    "lint:prettier": "prettier \"{src,e2e,examples}/*.{ts,tsx,js,jsx,css}\" \"**/*.{md,mdx,yml}\"",
```
ThisIsMissEm, line 12: I think we should really do this from the eslint and prettier configuration files, rather than via the command line options here, we should just be able to run `eslint` and it'd execute as above, same with prettier.

.eslintrc.js
```diff
@@ -7,15 +7,21 @@ module.exports = {
     "import/extensions": [".mjs", ".js"],
     "import/resolver": {
       node: {
-        extensions: [".mjs", ".ts"],
+        extensions: [".mjs", ".ts", ".js", ".jsx"],
+        moduleDirectory: [
+          "src",
+          "node_modules",
+          "e2e/**/node_modules",
+          "examples/**/node_modules",
```
ThisIsMissEm, line 15: These last two should be in overrides, for each specific subtree, so one for e2e and one for examples.

ThisIsMissEm (approved):
A few comments, but overall looks good. I don't know if we want to address those comments immediately though, but we might.

PR #185, opened 2022-03-30
Chore: update license header

ThisIsMissEm:
This updates the license header and fixes a few issues noticed whilst doing that. Note: for `e2e/*` the header was manually changed using a find & replace, because the `*.js` and `*.jsx` files are excluded from eslint at the moment due to tsconfig issues.

I've also added a few comments on things that might raise questions.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,5 @@ jest.e2e.config.js
 license-header.js
 rollup.config.js
 tsconfig.json
+# The only check that applies here is really the license header:
+.eslintrc.js
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,24 +1,3 @@
-/**
- * Copyright 2022 Inrupt Inc.
- *
- * Permission is hereby granted, free of charge, to any person obtaining a copy
- * of this software and associated documentation files (the "Software"), to deal in
- * the Software without restriction, including without limitation the rights to use,
- * copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the
- * Software, and to permit persons to whom the Software is furnished to do so,
- * subject to the following conditions:
- *
- * The above copyright notice and this permission notice shall be included in
- * all copies or substantial portions of the Software.
- *
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
- * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
- * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
- * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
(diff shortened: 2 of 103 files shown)


.eslintignore
```diff
@@ -4,3 +4,5 @@ jest.e2e.config.js
 license-header.js
 rollup.config.js
 tsconfig.json
+# The only check that applies here is really the license header:
+.eslintrc.js
```
ThisIsMissEm, line 8: As this file was a `.js` file, we weren't actually passing it through eslint anyway, however, it still complained about the license header missing, hence, exclude the configuration file from eslint itself.

examples/grant-access/src/grant-access.ts
```diff
@@ -1,29 +1,29 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
 
 /* eslint-disable no-console */
 
+// eslint-disable-next-line import/no-unresolved
 import { Session } from "@inrupt/solid-client-authn-node";
 import { config } from "dotenv-flow";
-// eslint-disable-next-line import/no-unresolved
```
ThisIsMissEm, line 26: See the eslint overrides

src/manage/redirectToRequestor.test.ts
```diff
@@ -1,23 +1,25 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+// PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+// HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+
+// This rule complains about the `@jest/globals` variables overriding global vars:
```
ThisIsMissEm, line 22: eslint had a [tissy](https://en.wiktionary.org/wiki/tissy) about import, so I overrode it with the same import from another file, and then it was like "all good", so I've no idea what was up with that.

.eslintrc.js
```diff
@@ -32,6 +11,14 @@ module.exports = {
       },
     },
   },
+  overrides: [
```
ThisIsMissEm, line 14: @NSeydoux this is what I was thinking for the e2e tests over on #184 — I've not actually done this for the e2e though, but we could.

ajacksified (approved):

.eslintrc.js
NSeydoux: Ah right, it's cleaner than individually muting the offending imports. I'm still a bit puzzled why this is required at all, but it is an improvement.

NSeydoux (approved):

PR #189, opened 2022-04-04
Denied access grants are access grants too

NSeydoux:
The access grant type guard was too strict, and enforced the access grant status only to be accepted, while denied is also valid.

### DIFF
--- a/src/guard/isAccessGrant.ts
+++ b/src/guard/isAccessGrant.ts
@@ -19,7 +19,10 @@
 // SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
-import { GC_CONSENT_STATUS_EXPLICITLY_GIVEN } from "../constants";
+import {
+  GC_CONSENT_STATUS_DENIED,
+  GC_CONSENT_STATUS_EXPLICITLY_GIVEN,
+} from "../constants";
 import {
   AccessGrantBody,
   BaseAccessVcBody,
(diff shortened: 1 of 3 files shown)


src/guard/isAccessGrant.ts
```diff
@@ -30,8 +33,9 @@ export function isAccessGrant(
 ): vc is BaseAccessVcBody & AccessGrantBody {
   return (
     (vc as AccessGrantBody).credentialSubject.providedConsent !== undefined &&
-    (vc as AccessGrantBody).credentialSubject.providedConsent.hasStatus ===
-      GC_CONSENT_STATUS_EXPLICITLY_GIVEN &&
+    [GC_CONSENT_STATUS_EXPLICITLY_GIVEN, GC_CONSENT_STATUS_DENIED].includes(
+      (vc as AccessGrantBody).credentialSubject.providedConsent.hasStatus
+    ) &&
```
ThisIsMissEm, line 38: What about access grants that are in other statuses? Perhaps it'd be better to be able to provide a range of statuses accepted when validating an AccessGrant?

ThisIsMissEm (changes requested):

src/guard/isAccessGrant.ts
```diff
@@ -19,7 +19,10 @@
 // SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
-import { GC_CONSENT_STATUS_EXPLICITLY_GIVEN } from "../constants";
+import {
+  GC_CONSENT_STATUS_DENIED,
```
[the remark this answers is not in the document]
NSeydoux: Good catch! However, we need to be aligned with the server even more so than with the vocabulary, so I'll loop this back to the ESS team and we'll see how to migrate from there.

src/guard/isAccessGrant.ts
NSeydoux: The intention here was to support only grants that are valid as per the shapes of the server-side consent service, which are currently only these two. Since this is a server-controlled thing, I'm not sure we want consumers to have freedom there.

matthieubosquet (approved):

ajacksified (approved):

ajacksified (approved):

PR #200, opened 2022-04-09
chore: fully remove testcafe as we've migrated to playwright

ThisIsMissEm:
Looks like a few testcafe related things were left over from the migration to playwright, this removes those dependencies and files.


package.json
```diff
@@ -102,14 +101,13 @@
     "fast-check": "^2.17.0",
     "jest": "^27.5.1",
     "license-checker": "^25.0.1",
+    "nlfurniss-typedoc-plugin-sourcefile-url": "^2.0.0",
```
ThisIsMissEm, line 104: This movement makes sense.. I'd probably manually added it as a dependency instead of `npm install --save-dev`

NSeydoux (approved):

PR #211, opened 2022-04-20
Fix: docs build issue due to myst-parser

ThisIsMissEm:
This repeats the change I just did on other repos.


NSeydoux (approved):

NSeydoux:
The `--keep-going` option sounds like cheering the parser up, I find it adorable.

PR #212, opened 2022-04-21
Refactor: remove unsupported access modes

ThisIsMissEm:
# New feature description

Removes the unsupported Access Modes for Access Grants. 

Also:
- stops publishing docs to gh-pages (somehow I missed this)
- stops outputting code coverage as files (we switched to text output everywhere else)
- switches to using `file:` instead of `npm link` for e2e tests
- fixes some dependency issues in the e2e test app
- renames the directory containing the e2e test app to "testApp"

# Checklist

- [ ] New functions/types have been exported in `index.ts`, if applicable.


.eslintignore
```diff
@@ -6,3 +6,4 @@ rollup.config.js
 tsconfig.json
 # The only check that applies here is really the license header:
 .eslintrc.js
+e2e/browser/testApp
```
ThisIsMissEm, line 9: @NSeydoux this is the only possible strategy I have here, as the e2e browser testApp's eslint configuration conflicts with the repos eslint config, which in turn conflicts with jest 27. I'm not sure how you were able to use Jest 27 before, as we've _never_ supported it.

e2e/browser/testApp/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": ["@inrupt/eslint-config-lib", "next/core-web-vitals"]
```
ThisIsMissEm, line 2: This is all we really should have, but this doesn't necessarily work because the testApp depends on prism & lit which are horribly out of date and really shouldn't be being used in e2e tests.

e2e/browser/testApp/components/appContent/index.jsx
```diff
@@ -87,7 +87,7 @@ export default function Home() {
       return;
     }
     (async () => {
-      const accessGrant = await approveAccessRequest(
+      const accessGrantRequest = await approveAccessRequest(
```
NSeydoux, line 90: Approving the request issues the actual access grant, so I'd keep the initial naming.

NSeydoux (approved):

e2e/browser/testApp/components/appContent/index.jsx
ThisIsMissEm: This change was because `accessGrant` is declared in the upper scope (no-shadow)

PR #240, opened 2022-05-31
Fixup demo app and export all types

matthieubosquet:


CHANGELOG.md
```diff
@@ -6,9 +6,14 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
 
+### Bugfix
+
+- Export all types via Rollup.
+- Fix demo application.
+
 ### New features
 
--
+- Drop Node 12 support.
```
ThisIsMissEm, line 16: This is also a breaking change, not a new feature.

matthieubosquet: Marked as such.

.github/workflows/cd.yml
```diff
@@ -136,7 +136,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        node-version: [16.x, 14.x, 12.x]
+        node-version: ["18", "16", "14"]
```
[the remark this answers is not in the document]
matthieubosquet: Removed

examples/grant-access/views/grant-form.ejs
```diff
@@ -0,0 +1,17 @@
...
+            <label for="deny">Deny</label>
+            <input type="radio" name="manage" id="deny" value="deny" required />
+            <input id="requestVc" type="hidden" name="requestVc" value='<%= JSON.stringify(accessRequest) %>'>
+            <input id="redirectUrl" type="hidden" name="redirectUrl" value="<%- requestorRedirectUrl %>">
+        </p>
+        <p>
+            <input type="submit" value="Respond to Access Request">
+        </p>
+    </fieldset>
+</form>
```
ThisIsMissEm, line 17: missing newline at EOF

src/index.ts
```diff
@@ -19,18 +19,22 @@
...
 
-export type { AccessModes } from "./type/AccessModes";
 export type { AccessBaseOptions } from "./type/AccessBaseOptions";
+export type { AccessCredentialType } from "./type/AccessCredentialType";
+export type { AccessGrantContext } from "./type/AccessGrantContext";
+export type { AccessModes } from "./type/AccessModes";
+export type { ApproveAccessRequestOverrides } from "./manage/approveAccessRequest";
+export type { FetchOptions } from "./type/FetchOptions";
 export type { IssueAccessRequestParameters } from "./type/IssueAccessRequestParameters";
+export type { RedirectOptions } from "./type/RedirectOptions";
```
ThisIsMissEm, line 29: Might be good to merge this type into `RedirectToAccessManagementUiOptions` rather than having it seperate?

ThisIsMissEm (approved):
I'm not confident on the changes made to the root tsconfig files, especially changing the target from es2015/2017 to es2020; I think we do still want to output for es2015 for browsers.

I'm going to approve, but it'd be better to make those changes in a more scheduled fashion when we've time to reason about what the implications are.

tsconfig.json
```diff
@@ -1,23 +1,20 @@
 {
   "compilerOptions": {
     /* Basic Options */
-    "target": "es2017",
+    "target": "es2020",
```
matthieubosquet, line 4:
```suggestion
    "target": "es2017",
```

examples/grant-access/views/grant-form.ejs
matthieubosquet:
```suggestion
</form>

```

src/index.ts
matthieubosquet: Sounds good, but maybe let's address this in a separate PR.

ThisIsMissEm: sure!

matthieubosquet:
> I'm not confident on the changes made to the root tsconfig files, especially changing the target from es2015/2017 to es2020; I think we do still want to output for es2015 for browsers.
>
> I'm going to approve, but it'd be better to make those changes in a more scheduled fashion when we've time to reason about what the implications are.

I reverted to target `ES2017` as it was before. My general idea would be to target the maintenance Node LTS (14 is ES2020) and let parcel/browserify & co deal with older browser requirements.

ThisIsMissEm:
> > I'm not confident on the changes made to the root tsconfig files, especially changing the target from es2015/2017 to es2020; I think we do still want to output for es2015 for browsers.
> > I'm going to approve, but it'd be better to make those changes in a more scheduled fashion when we've time to reason about what the implications are.
> 
> I reverted to target `ES2017` as it was before. My general idea would be to target the maintenance Node LTS (14 is ES2020) and let parcel/browserify & co deal with older browser requirements.

Yup, but that's something we can do in a planned manner, not as a small change in a larger PR, especially this close to a big release.

PR #255, opened 2022-06-13
fix: issue with dependency conflicts

ThisIsMissEm:


matthieubosquet (approved):

ajacksified (changes requested):
Let's bump rather than applying overrides.

PR #260, opened 2022-06-17
Instanciate VC context based on issuer domain

NSeydoux:
According to the [ESS documentation](https://docs.inrupt.com/ess/latest/services/service-vc/#ess-vc-service-endpoints), the JON-LD context for issued VCs will depend on the issuer's domain. This adds a function to instantiate the context properly under these assumptions.

### DIFF
--- a/src/guard/isAccessGrantContext.ts
+++ b/src/guard/isAccessGrantContext.ts
@@ -19,9 +19,9 @@
 // SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
-import { ACCESS_GRANT_CONTEXT } from "../constants";
+import { ACCESS_GRANT_CONTEXT_DEFAULT } from "../constants";
 import { AccessGrantContext } from "../type/AccessGrantContext";
 
 export function isAccessGrantContext(x: unknown): x is AccessGrantContext {
-  return Array.isArray(x) && ACCESS_GRANT_CONTEXT.every((y) => x.includes(y));
+  return Array.isArray(x) && ACCESS_GRANT_CONTEXT_DEFAULT.every((y) => x.includes(y));
 }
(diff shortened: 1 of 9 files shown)


src/guard/isAccessGrantContext.ts
```diff
@@ -19,9 +19,9 @@
 // SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
-import { ACCESS_GRANT_CONTEXT } from "../constants";
+import { ACCESS_GRANT_CONTEXT_DEFAULT } from "../constants";
 import { AccessGrantContext } from "../type/AccessGrantContext";
 
 export function isAccessGrantContext(x: unknown): x is AccessGrantContext {
-  return Array.isArray(x) && ACCESS_GRANT_CONTEXT.every((y) => x.includes(y));
+  return Array.isArray(x) && ACCESS_GRANT_CONTEXT_DEFAULT.every((y) => x.includes(y));
```
matthieubosquet, line 26:
```suggestion
  return Array.isArray(x) && x.includes(CONTEXT_VC_W3C);
```

src/guard/isAccessGrantContext.ts
```diff
@@ -19,9 +19,9 @@
 // SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
-import { ACCESS_GRANT_CONTEXT } from "../constants";
+import { ACCESS_GRANT_CONTEXT_DEFAULT } from "../constants";
```
matthieubosquet, line 22:
```suggestion
import { CONTEXT_VC_W3C } from "../constants";
```

src/guard/isAccessGrantContext.ts
matthieubosquet: This ultimately is also used by `getAccessGrant` and `approveAccessRequest` so we can't be too picky in the type guard. I don't think it is worth being more specific for now.

matthieubosquet (approved):
Just a little issue with the type guard but otherwise looks good! :)

src/guard/isAccessGrantContext.ts
NSeydoux: Fixed in c186802

PR #273, opened 2022-06-27
Remove unnecessary type casting in demo apps

matthieubosquet:
This PR removes a [now unnecessary type cast](https://github.com/inrupt/solid-client-authn-js/releases/tag/v1.12.1).


ThisIsMissEm (approved):

ThisIsMissEm:
This closes #274 right?

NSeydoux:
I don't think so: this only updates the Node libraries and their dependencies, and #274 is for the browser library

ThisIsMissEm:
Ah, okay.

PR #276, opened 2022-06-28
Upgrade typescript & shared eslint configuration

NSeydoux:
This upgrades to the latest version of typescript, and fixes the mismatch in the fetch function signature it involves.


matthieubosquet (approved):
Updating TypeScript is fine, but you mixed it up with https://github.com/inrupt/solid-client-access-grants-js/pull/277/files.

Maybe remove the issueAccess* changes before merging?

NSeydoux:
Yeah #277 was required to pass e2e tests. I was planing on merging it first, and then rebase this on top.

package.json
```diff
@@ -81,8 +81,8 @@
   "devDependencies": {
     "@inrupt/eslint-config-base": "^1.0.0",
     "@inrupt/eslint-config-lib": "^1.0.0",
```
ThisIsMissEm, line 83:
```suggestion
    "@inrupt/eslint-config-lib": "^1.0.0",
```

You no longer need `base` here.

ThisIsMissEm: Actually, ignore that. This repo hasn't been correctly upgraded to 1.0.0

ThisIsMissEm (changes requested):
@NSeydoux I'm going to take this over, as we need to do the correct upgrade to the new eslint configs, so I'll push a commit with the changes required for that.

ThisIsMissEm:
Okay, the Vercel crash here is related to the sourceurl plugin and typedoc 0.23.3, due to this breaking change:
> Removed ProjectReflection.files, this was an internal cache that should not have been exposed, and shouldn't have existed in the first place, since removing it made TypeDoc faster.

I'm going to remove the sourceurl plugin for now, as it was a "nice to have" on Vercel, but worked out of the box on the published docs.

ThisIsMissEm (approved):
I think this is now good to go.

PR #277, opened 2022-06-28
Do not include default VC domain in request

NSeydoux:
The default fetch is bound to the default Inrupt VC domain, and we don't want to include it in all VCs.

This fixes an issue introduced in the previous commit, which was automerged by mistake because e2e tests were not gating (which they are now).


ThisIsMissEm (approved):
This looks okay to me.

PR #278, opened 2022-06-28
Accept URL object

NSeydoux:
Retrieving data from redirect URLs is a pattern this library relies on. In some cases, the URL being passed will have been parsed first, depending on the framework used, so accepting URL objects should make the API slightly friendlier to some use cases.

### DIFF
--- a/src/manage/getAccessRequestFromRedirectUrl.test.ts
+++ b/src/manage/getAccessRequestFromRedirectUrl.test.ts
@@ -135,9 +135,10 @@ describe("getAccessRequestFromRedirectUrl", () => {
     ) as jest.Mocked<{
       getVerifiableCredential: typeof getVerifiableCredential;
     }>;
-    vcModule.getVerifiableCredential.mockResolvedValueOnce(
-      mockAccessRequestVc()
-    );
+    vcModule.getVerifiableCredential
+      .mockResolvedValueOnce(mockAccessRequestVc())
+      // We test this twice, once for string and one for URL object.
+      .mockResolvedValueOnce(mockAccessRequestVc());
 
     const redirectedToUrl = new URL("https://redirect.url");
     redirectedToUrl.searchParams.append(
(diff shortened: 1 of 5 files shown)


ThisIsMissEm (changes requested):
This is great, but we need to have test coverage for passing both a string and a URL object.

src/manage/getAccessRequestFromRedirectUrl.test.ts
```diff
@@ -155,7 +155,7 @@ describe("getAccessRequestFromRedirectUrl", () => {
     );
 
     const { accessRequest, requestorRedirectUrl } =
-      await getAccessRequestFromRedirectUrl(redirectedToUrl.href);
+      await getAccessRequestFromRedirectUrl(redirectedToUrl);
```
NSeydoux, line 158: @ThisIsMissEm that's a bit far fetched, but this actually does add coverage for using an URL object instead of a plain string. However I think you mentioning it had to be done indicates that this is too implicit, which I had kind of a feeling when doing it in the first place. There is some setup in the tests I didn't want to duplicate, but I'll see how I can make this cleaner.

ThisIsMissEm: I think maybe you could just do:

```
const accessRequestFromString = await getAccessRequestFromRedirectUrl(redirectedToUrl.href);
const accessRequestFromUrl = await getAccessRequestFromRedirectUrl(redirectedToUrl);

expect(accessRequestFromString.accessRequest).toStrictEqual(accessRequestFromUrl.accessRequest);
expect(accessRequestFromString.requestorRedirectUrl).toBe(accessRequestFromUrl.requestorRedirectUrl);

const { accessRequest, requestorRedirectUrl } = accessRequestFromUrl;
```

src/manage/getAccessRequestFromRedirectUrl.test.ts
```diff
@@ -135,9 +135,10 @@ describe("getAccessRequestFromRedirectUrl", () => {
     ) as jest.Mocked<{
       getVerifiableCredential: typeof getVerifiableCredential;
     }>;
-    vcModule.getVerifiableCredential.mockResolvedValueOnce(
-      mockAccessRequestVc()
-    );
+    vcModule.getVerifiableCredential
+      .mockResolvedValueOnce(mockAccessRequestVc())
+      // We test this twice, once for string and one for URL object.
+      .mockResolvedValueOnce(mockAccessRequestVc());
```
ThisIsMissEm, line 141: Would've probably been nicer to just mockResolvedValue without the once.

ThisIsMissEm (commented):
One comment, but LGTMO

PR #319, opened 2022-08-30
Feature: additional File APIs

ThisIsMissEm:
# New feature description

This pull request adds the following APIs that mirror their solid-client counterparts:

- `overwriteFile`
- `saveFileInContainer`

It also:

- Improves end-to-end test coverage for all Resource APIs
- Cleans up some eslint comments which are now controlled from eslintrc.
- Updates the browserlist package (we kept getting warnings about this)
- Switches to using `URLSearchParams` instead of a customer serialiser for the Access Grant API request payload, URLSearchParams works in all environments we support, so this should be a safe change.

# Checklist


e2e/node/e2e.test.ts
```diff
@@ -303,4 +325,355 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
...
+      const updatedDatasetAsOwnerTtl = await solidClient.solidDatasetAsTurtle(
+        updatedDatasetAsOwner
+      );
+
+      // Assert that the dataset changed:
+      expect(updatedDatasetTtl).not.toBe(existingDatasetAsOwnerTtl);
+      expect(updatedDatasetTtl).toBe(updatedDatasetAsOwnerTtl);
+    });
+
+    it.skip("can use the saveSolidDatasetAt API for a new dataset", async () => {
```
ThisIsMissEm, line 485: @NSeydoux hey, I wasn't able to get this working as an e2e test, but we've had it as an API for a while. I'm not sure if this would ever work as expected, given the way these APIs are written.

ThisIsMissEm:
I've included some tech debt fixes in this pull request too, where to address an issue I made the change in `eslintrc` rather than line-comment disabling rules, so I had to go back through all the other files and clean up all those line comments. 

I've added the new methods and tried to add "okay" e2e test coverage, though haven't been able to with `saveFileInContainer` unless the file already exists (I get a 401 due to how that API works, but there's a work around which is to use `overwriteFile` in that case.

src/manage/approveAccessRequest.test.ts
```diff
@@ -19,10 +19,11 @@
 // SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
-/* eslint-disable no-shadow */
-import { mockSolidDatasetFrom } from "@inrupt/solid-client";
 import { jest, it, describe, expect } from "@jest/globals";
+
```
jeswr, line 23: Unecessary line break

src/resource/file.ts
```diff
@@ -0,0 +1,157 @@
...
+): ReturnType<typeof coreGetFile> {
+  const fetchOptions: FetchOptions = {};
+  if (options && options.fetch) {
+    fetchOptions.fetch = options.fetch;
+  }
+
+  const authenticatedFetch = await fetchWithVc(
+    resourceUrl,
+    accessGrant,
+    fetchOptions
```
jeswr, line 55: Why not just 

```ts
const authenticatedFetch = await fetchWithVc(
    resourceUrl,
    accessGrant,
    { fetch: options?.fetch }
```

The current approach feels verbose and takes a few seconds to digest.

src/resource/file.ts
```diff
@@ -0,0 +1,157 @@
...
+  );
+
+  return coreGetFile.apply(undefined, [
```
jeswr, line 58: Is there any benefit to using `.apply` over just calling `coreGetFile` directly in this instance - I can't see one.

src/resource/file.ts
```diff
@@ -0,0 +1,157 @@
...
+  const fetchOptions: FetchOptions = {};
+  if (options && options.fetch) {
+    fetchOptions.fetch = options.fetch;
+  }
+
+  const authenticatedFetch = await fetchWithVc(
+    resourceUrl,
+    accessGrant,
+    fetchOptions
+  );
```
jeswr, line 94: again with the `fetchOptions`

src/resource/file.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  return coreOverwriteFile.apply(undefined, [
+    resourceUrl,
+    file,
+    {
+      ...overwriteFileOptions,
+      fetch: authenticatedFetch,
+    },
+  ]);
```
jeswr, line 108: Same comment as `fetchOptions` for `overwriteFileOptions`.
Same comment for `.apply`

src/resource/file.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  return coreSaveFileInContainer.apply(undefined, [
+    containerUrl,
+    file,
+    {
+      ...fileOptions,
+      fetch: authenticatedFetch,
+    },
+  ]);
+}
```
jeswr, line 157: same comments for `fileOptions` and `.apply`

src/resource/file.ts
```diff
@@ -0,0 +1,157 @@
...
+    {
+      fetch: authenticatedFetch,
+    },
+  ]);
+}
+
+/**
+ * Saves a file at a given URL using an Access Grant to prove the caller is
+ * authorized to save a file at the given URL. If a file already exists in the
+ * at the URL, the function overwrites the existing file.
```
jeswr, line 69:
```suggestion
 * authorized to save a file at the given URL. If a file already exists
 * at the URL, the function overwrites the existing file.
```

src/resource/getSolidDataset.ts
```diff
@@ -43,14 +42,22 @@ export async function getSolidDataset(
...
   );
 
-  return wrapSolidClientFunction(coreGetSolidDataset, [datasetUrl], {
-    ...options,
-    fetch: authenticatedFetch,
-  });
+  return coreGetSolidDataset.apply(undefined, [
+    datasetUrl,
+    {
+      ...options,
```
jeswr, line 59: ditto on the `fetchOptions`

src/resource/file.ts
jeswr, line 69: I would reccomend copy/pasting the addtional docs from the corresponding functions in `solid-client-js`. For instance this documentation is missing the description of defaulting behavior around `content-type`.

jeswr (changes requested):

src/resource/file.ts
ThisIsMissEm: I don't want to be duplicating docs between repos, but perhaps we can do a "see also" link, I'd have to check with typedocs documentation.

src/resource/file.ts
ThisIsMissEm: This results in `fetch` being undefined:

Screenshot 2022-08-31 at 16 19 29

Which likely results in the method not working correctly. I would see this all as temporary though, until we release the major making fetch passing mandatory.

src/manage/approveAccessRequest.test.ts
ThisIsMissEm: Was actually deliberate, as it's kinda a "good to do" to have the `@jest/globals` line be the first line in the file, and also separate from dependencies used for testing

src/resource/file.ts
ThisIsMissEm: I think the key benefit might come from it appearing in stracktraces through `.apply`, iirc, where as just a `return call(` might not list `call` in the stacktrace. Though really we should probably be doing `return await coreGetFile(` instead.

jeswr (approved):

PR #323, opened 2022-08-31
WIP: Refactor e2e tests

ThisIsMissEm:
This is a branch that will probably never be merged, but carries an idea for refactoring our e2e tests. Currently we've a lot of repeated code in the e2e tests, and we don't always teardown the resources we create during a test run (this can lead to a pod becoming excessively large which has caused an issue in the past).

The approach here is two fold:

1) Create some shared utilities that can be used across e2e test files.
2) Split the e2e tests up into separate files

One of the utilities I've created is this `TestResourceTracker` class, where it manages all the IRIs you work with throughout your tests, such that they can be cleaned up when the tests finish.


e2e/node/e2e.test.ts
```diff
@@ -19,7 +19,6 @@
 // SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
```
ThisIsMissEm, line 19: You can ignore the changes in this file, it's not relevant.

NSeydoux:
This branch is outdated, let's close this and reconsider if applicable in the future.

PR #354, opened 2022-10-03
No longer accept VCs being passed by value

NSeydoux:
VCs were already no longer passed by value, and only by IRI, but now a VC being passed by value will be ignored, and not picked up.

(Would it makes sense to add a test to verify that the legacy parameter is now ignored ? Seems a bit overkill)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Deprecation notice
+
+- `getAccessRequestFromRedirectUrl` and `getAccessGrantFromRedirectUrl` no longer
+  pick up VC passed by value, and only support passing by IRI. This is for security
+  reasons, as passing a VC by value in the IR leaks information.
+
 ### New features
 
 - `getAccessRequestFromRedirectUrl` and `getAccessGrantFromRedirectUrl` now accept URL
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Deprecation notice
+
+- `getAccessRequestFromRedirectUrl` and `getAccessGrantFromRedirectUrl` no longer
+  pick up VC passed by value, and only support passing by IRI. This is for security
+  reasons, as passing a VC by value in the IR leaks information.
```
ThisIsMissEm, line 11: "in the IR"? Is this a typo of IRI?

docs/api/requirements.txt
```diff
@@ -1,5 +1,5 @@
 # pip install -r requirements.txt
 
-pydata-sphinx-theme==0.9.0
+pydata-sphinx-theme==0.10.1
```
ThisIsMissEm, line 3: Careful with this update, so far there's been some weird breaking changes in the python dependency updates — I'd suggest only updating to the same version as solid-client or notifications

ThisIsMissEm (approved):
Looks good, but let's make sure this is correctly labeled as a semver major.

CHANGELOG.md
NSeydoux: Yup. Fixed in ee9a9aae3a7c0187c24dc2b0f9d9e7b970e145e8

docs/api/requirements.txt
NSeydoux: Yeah, it has been taken care of by @kay-kim in https://github.com/inrupt/solid-client-access-grants-js/pull/320, and removed from the current PR.

PR #400, opened 2022-11-18
chore: enable matrix concurrency and migrate to typescript-sdk-tools

ThisIsMissEm:


.github/dependabot.yml
```diff
@@ -25,7 +25,7 @@ updates:
   # Enable version updates for embedded demo app
   - package-ecosystem: "npm"
     # Look for `package.json` and `lock` files in the `e2e/browser/webapp` directory.
-    directory: "/e2e/browser/webapp"
+    directory: "/e2e/browser/testApp"
```
NSeydoux, line 28: Good catch, can you update/delete the comment above?

.github/workflows/ci.yml
```diff
@@ -2,14 +2,21 @@ name: CI
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 jobs:
+  lint:
+    uses: inrupt/typescript-sdk-tools/.github/workflows/reusable-lint.yml@v1.4.2
+
   build:
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
```
NSeydoux, line 14: Are we setting this flag up for unit tests too?

.github/workflows/e2e-node.yml
```diff
@@ -14,12 +17,19 @@ jobs:
...
-        node-version: [16.x]
-        # NSS does not support static client registration, which we rely on for testing.
+        node-version: ["14.x", "16.x"]
+        # FIXME: deploy new environments:
+        # environment-name: ["ESS PodSpaces"]
         environment-name: ["ESS Dev-Next"]
+        experimental: [false]
+        # include:
+        #   - environment-name: "ESS Dev-Next"
+        #     experimental: true
```
NSeydoux, line 32:
```suggestion
        #     experimental: true
        #     node-version: 16.x
```

NSeydoux (approved):

.github/workflows/ci.yml
ThisIsMissEm: That's because this is running the unit tests in a matrix across node.js 16 and 14, not just the E2E tests.

ThisIsMissEm: If we include 18, then we'd have 18 as experimental

.github/dependabot.yml
```diff
@@ -25,7 +25,7 @@ updates:
   # Enable version updates for embedded demo app
   - package-ecosystem: "npm"
     # Look for `package.json` and `lock` files in the `e2e/browser/webapp` directory.
```
ThisIsMissEm, line 27:
```suggestion
    # Look for `package.json` and `lock` files in the `e2e/browser/testApp` directory.
```

.github/workflows/ci.yml
NSeydoux: Well I'm not sure: when we move to supporting Node 18, is it going to be considered experimental? In other words, are any of our unit tests experimental at some point, or does it only apply to e2e tests, where the remote environment is experimental rather than our runtime?

ThisIsMissEm: Short answer is: I don't know.

PR #404, opened 2022-11-22
feat: Add createContainerInContainer method

solid-akb:
# Add CreateContainerInContainer method
This PR adds the createContainerInContainer function in the Access Grants library. 
The ticket for this is [SDK-2791](https://inrupt.atlassian.net/browse/SDK-2791)

Why create container in container? Here's a little reminder of the use case:

This function is primarily useful if the current user does not have access to change existing files in a Container, but is allowed to add new files; in other words, they have Append, but not Write access to a Container. This is useful in situations where someone wants to allow others to, for example, send notifications to their Pod, but not to view or delete existing notifications.

# Checklist

- [x] New modules (i.e. new `.ts` files) are listed in the `exports` field in `package.json`, if applicable.
- [x] New modules (i.e. new `.ts` files) are listed in the `typedocOptions.entryPoints` field in `tsconfig.json`, if applicable.

### DIFF
--- a/e2e/node/e2e.test.ts
+++ b/e2e/node/e2e.test.ts
@@ -35,17 +35,18 @@ import { Session } from "@inrupt/solid-client-authn-node";
 import { isVerifiableCredential } from "@inrupt/solid-client-vc";
 import { getNodeTestingEnvironment } from "@inrupt/internal-test-env";
 import {
-  approveAccessRequest,
-  getAccessGrantAll,
-  isValidAccessGrant,
-  issueAccessRequest,
-  revokeAccessGrant,
-  getFile,
   AccessGrant,
+  approveAccessRequest,
+  createContainerInContainer,
+  getAccessGrantAll,
+  getFile,
+  getSolidDataset,
+  issueAccessRequest,
+  isValidAccessGrant,
   overwriteFile,
+  revokeAccessGrant,
   saveFileInContainer,
   saveSolidDatasetAt,
-  getSolidDataset,
 } from "../../src/index";
 
 const env = getNodeTestingEnvironment({
@@ -378,10 +379,10 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
     let testFileName: string;
     let testFileIri: string;
     let testContainerIri: string;
+    let testContainerIriChild: string;
 
     beforeEach(async () => {
       const containerPath = `${resourceOwnerSession.info.sessionId}-dataset-apis`;
-
       testContainerIri = new URL(`${containerPath}/`, resourceOwnerPod).href;
       testFileName = "dataset.ttl";
       testFileIri = new URL(testFileName, testContainerIri).href;
@@ -440,6 +441,10 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
       });
 
       await solidClient.deleteFile(testFileIri, {
+        fetch: resourceOwnerSession.fetch,
+      });
+
+      await solidClient.deleteContainer(testContainerIriChild, {
         fetch: resourceOwnerSession.fetch,
       });
 
@@ -517,6 +522,30 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
       // Assert that the dataset changed:
       expect(updatedDatasetTtl).not.toBe(existingDatasetAsOwnerTtl);
       expect(updatedDatasetTtl).toBe(updatedDatasetAsOwnerTtl);
+    });
+
+    it("can use the createContainerInContainer API to create a new container", async () => {
+      const containerNameSuggestion = "newTestContainer";
+      testContainerIriChild = `${testContainerIri}${containerNameSuggestion}/`;
+      await createContainerInContainer(testContainerIri, accessGrant, {
+        fetch: requestorSession.fetch,
+        slugSuggestion: containerNameSuggestion,
+      });
+      const parentContainer = await solidClient.getSolidDataset(
+        testContainerIri,
+        {
+          fetch: resourceOwnerSession.fetch,
+        }
+      );
+
+      const parentContainerContainsAll =
+        solidClient.getThingAll(parentContainer);
+
+      const match = parentContainerContainsAll.filter((thing) => {
+        return thing.url === testContainerIriChild;
+      });
+
+      expect(match).toHaveLength(1);
     });
 
     it.skip("can use the saveSolidDatasetAt API for a new dataset", async () => {
--- a/e2e/node/e2e.test.ts
+++ b/e2e/node/e2e.test.ts
@@ -524,7 +524,7 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
       expect(updatedDatasetTtl).toBe(updatedDatasetAsOwnerTtl);
     });
 
-  it("can use the createContainerInContainer API to create a new container", async () => {
+    it("can use the createContainerInContainer API to create a new container", async () => {
       const containerNameSuggestion = "newTestContainer";
       const newChildContainer = await createContainerInContainer(
         testContainerIri,
@@ -551,7 +551,7 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
 
       expect(match).toHaveLength(1);
     });
-    
+
     it("can use the saveSolidDatasetInContainer API for an existing dataset", async () => {
       // Need to delete dataset that was already created in test setup,
       // such that our test can create an empty dataset at `testFileIri`.
--- a/e2e/node/e2e.test.ts
+++ b/e2e/node/e2e.test.ts
@@ -29,10 +29,12 @@ import {
   beforeEach,
   afterEach,
 } from "@jest/globals";
-import * as solidClient from "@inrupt/solid-client";
 import { Session } from "@inrupt/solid-client-authn-node";
 import { isVerifiableCredential } from "@inrupt/solid-client-vc";
 import { getNodeTestingEnvironment } from "@inrupt/internal-test-env";
+// Making a named import here to avoid confusion with the wrapped functions from
+// the access grant API
+import * as sc from "@inrupt/solid-client";
 import {
   AccessGrant,
   approveAccessRequest,
@@ -94,7 +96,7 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
     });
 
     // Create a file in the resource owner's Pod
-    const resourceOwnerPodAll = await solidClient.getPodUrlAll(
+    const resourceOwnerPodAll = await sc.getPodUrlAll(
       resourceOwnerSession.info.webId as string
     );
     if (resourceOwnerPodAll.length === 0) {
@@ -105,7 +107,7 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
     // eslint-disable-next-line prefer-destructuring
     resourceOwnerPod = resourceOwnerPodAll[0];
 
-    const savedFile = await solidClient.saveFileInContainer(
+    const savedFile = await sc.saveFileInContainer(
       resourceOwnerPodAll[0],
       Buffer.from(SHARED_FILE_CONTENT),
       {
@@ -115,14 +117,14 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
       }
     );
 
-    sharedFileIri = solidClient.getSourceUrl(savedFile);
+    sharedFileIri = sc.getSourceUrl(savedFile);
   });
 
   // Cleanup the shared file
   afterAll(async () => {
     if (sharedFileIri) {
       // Remove the shared file from the resource owner's Pod.
-      await solidClient.deleteFile(sharedFileIri, {
+      await sc.deleteFile(sharedFileIri, {
         fetch: resourceOwnerSession.fetch,
       });
     }
@@ -387,24 +389,21 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
       testFileName = "dataset.ttl";
       testFileIri = new URL(testFileName, testContainerIri).href;
 
-      await solidClient.createContainerAt(testContainerIri, {
+      await sc.createContainerAt(testContainerIri, {
         fetch: resourceOwnerSession.fetch,
       });
 
-      const newThing = solidClient.setStringNoLocale(
-        solidClient.createThing({
+      const newThing = sc.setStringNoLocale(
+        sc.createThing({
           name: "e2e-test-thing",
         }),
         "https://arbitrary.vocab/regular-predicate",
         "initial-dataset"
       );
 
-      const dataset = solidClient.setThing(
-        solidClient.createSolidDataset(),
-        newThing
-      );
+      const dataset = sc.setThing(sc.createSolidDataset(), newThing);
 
-      await solidClient.saveSolidDatasetInContainer(testContainerIri, dataset, {
+      await sc.saveSolidDatasetInContainer(testContainerIri, dataset, {
         fetch: resourceOwnerSession.fetch,
         slugSuggestion: testFileName,
       });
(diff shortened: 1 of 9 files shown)


solid-akb:
My first SDK PR! I have a few questions: 
This PR will require a change to the Change Log.  Any rules/guidelines on changing the Change Log?

src/resource/createContainerInContainer.ts
```diff
@@ -0,0 +1,81 @@
...
+  UrlString,
+  createContainerInContainer as coreCreateContainerInContainer,
+  WithResourceInfo,
+} from "@inrupt/solid-client";
+import { VerifiableCredential } from "@inrupt/solid-client-vc";
+import { fetchWithVc } from "../fetch";
+import { FetchOptions } from "../type/FetchOptions";
+
+/**
+ * Create an empty Container inside the Container at the given URL.
```
solid-akb, line 32: Can someone review the documentation here. I just copied it from solid-client-js and I don't _think_ it would differ but would love extra eyes on this to confirm.

src/resource/createContainerInContainer.ts
```diff
@@ -0,0 +1,81 @@
...
+ * for example, send notifications to their Pod, but not to view or delete existing notifications.
+ * You can pass a suggestion for the new Resource's name, but the server may decide to give it
+ * another name — for example, if a Resource with that name already exists inside the given
+ * Container.
+ * If the user does have access to write directly to a given location, [[createContainerAt]]
+ * will do the job just fine, and does not require the parent Container to exist in advance.
+ *
+ * @param containerUrl URL of the Container in which the empty Container is to be created.
+ * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters). `options.slugSuggestion` accepts a string for your new Container's name.
+ * @returns A promise that resolves to a SolidDataset if successful, and that
```
solid-akb, line 51: The return value wasn't included in solid-client-js. I added it here and will open a new PR to add it to the client-js docs. I think this is just a small documentation bug but let me know if it was left out intentionally.

src/resource/createContainerInContainer.ts
```diff
@@ -0,0 +1,81 @@
...
+ * rejects otherwise.
+ * @since 0.2.0
+ */
+
+type SaveInContainerOptions = Partial<
+  FetchOptions & {
+    slugSuggestion: string;
+  }
+>;
+export async function createContainerInContainer<SolidDataset>(
```
solid-akb, line 61: I noticed in the [saveSolidDatasetAt](https://github.com/inrupt/solid-client-access-grants-js/blob/7bd8c12c35289759c87f2898bc105dca9bc44be6/src/resource/saveSolidDatasetAt.ts#L50) we extend `Dataset extends SolidDataset` . Is this necessary? I took that out here as it felt like it was adding an unnecessary refinement.

src/resource/createContainerInContainer.ts
```diff
@@ -0,0 +1,81 @@
...
+ * access to a Container. This is useful in situations where someone wants to allow others to,
+ * for example, send notifications to their Pod, but not to view or delete existing notifications.
+ * You can pass a suggestion for the new Resource's name, but the server may decide to give it
+ * another name — for example, if a Resource with that name already exists inside the given
+ * Container.
+ * If the user does have access to write directly to a given location, [[createContainerAt]]
+ * will do the job just fine, and does not require the parent Container to exist in advance.
+ *
+ * @param containerUrl URL of the Container in which the empty Container is to be created.
+ * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters). `options.slugSuggestion` accepts a string for your new Container's name.
```
solid-akb, line 50: I added the bit about the `slugSuggestion` as that felt like an important thing to highlight. Happy to remove if it feels unnecessary or inconsistent with other documentation.

e2e/node/e2e.test.ts
```diff
@@ -35,17 +35,18 @@ import { Session } from "@inrupt/solid-client-authn-node";
 import { isVerifiableCredential } from "@inrupt/solid-client-vc";
 import { getNodeTestingEnvironment } from "@inrupt/internal-test-env";
 import {
+  AccessGrant,
```
solid-akb, line 38: There are some sorting changes here. If we keep the lists sorted, and we don't just add new imports to the end of an import list, it minimizes the chance for a merge conflicts. And it looks nice :)

src/resource/createContainerInContainer.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+describe("createContainerInContainer", () => {
+  it("authenticates using the provided VC", async () => {
+    const solidClientModule = jest.requireMock("@inrupt/solid-client") as any;
+    const mockedDataset = mockSolidDatasetFrom("https://some.url");
+    solidClientModule.createContainerInContainer.mockResolvedValueOnce(
+      MOCKED_DATASET
+    );
+    const mockedFetch = jest.fn() as typeof fetch;
+    // TODO: change to mockAccessGrantVc when rebasing
```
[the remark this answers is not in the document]
ThisIsMissEm: This isn't needed in your work here. `mockAccessRequestVc` is actually the mock function you want, it's just poorly named.

src/index.test.ts
```diff
@@ -21,45 +21,47 @@
...
+  issueAccessRequest,
+  isValidAccessGrant,
+  odrl,
+  overwriteFile,
+  redirectToAccessManagementUi,
+  redirectToRequestor,
+  revokeAccessGrant,
+  saveFileInContainer,
+  saveSolidDatasetAt,
 } from "./index";
```
ThisIsMissEm, line 58: These imports aren't actually intended to be alphabetically sorted, instead they're more grouped by the functionality / section of the SDK — it's a little confusing, but it basically tries to keep the various parts of the SDK together as a block

src/resource/createContainerInContainer.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    const resultDataset = await createContainerInContainer(
+      TEST_CONTAINER_URL,
+      mockAccessRequestVc(),
+      { fetch: mockedFetch }
```
ThisIsMissEm, line 51: Missing the slug for what to call the child container

src/resource/createContainerInContainer.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    );
+    expect(fetchWithVc).toHaveBeenCalledWith(
+      expect.anything(),
```
ThisIsMissEm, line 54:
```suggestion
      TEST_CONTAINER_URL,
```

It's annoying that this is testing internals, but we'll fix it later.

src/resource/createContainerInContainer.ts
```diff
@@ -0,0 +1,82 @@
...
+
+interface SaveInContainerOptions extends FetchOptions {
+  slugSuggestion?: string;
+}
+
+export async function createContainerInContainer(
+  containerUrl: UrlString,
+  accessGrant: VerifiableCredential,
+  options?: SaveInContainerOptions
+): Promise<SolidDataset & WithResourceInfo> {
```
ThisIsMissEm, line 65:
```suggestion
) {
```

This is deliberate, as we're inferring the return type from `coreCreateContainerInContainer`

src/resource/createContainerInContainer.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  const authenticatedFetch = await fetchWithVc(
+    containerUrl,
+    accessGrant,
+    fetchOptions
+  );
+
+  return await coreCreateContainerInContainer(containerUrl, {
+    ...options,
```
ThisIsMissEm, line 79: Best to build up the options to pass along explicitly, as is done elsewhere

ThisIsMissEm (changes requested):
A few minor changes

e2e/node/e2e.test.ts
```diff
@@ -519,6 +524,30 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
       expect(updatedDatasetTtl).toBe(updatedDatasetAsOwnerTtl);
     });
 
+    it("can use the createContainerInContainer API to create a new container", async () => {
+      const containerNameSuggestion = "newTestContainer";
+      testContainerIriChild = `${testContainerIri}${containerNameSuggestion}/`;
```
NSeydoux, line 529:
```suggestion
      testContainerIriChild = new URL(`${containerNameSuggestion}/`, testContainerIri).href;
```
We usually try to build URLs using the native URL object rather than string concatenation, which is more error-prone. It doesn't change much in this context, but it's a general guideline in the codebase :)

e2e/node/e2e.test.ts
```diff
@@ -519,6 +524,30 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
...
+      });
+      const parentContainer = await solidClient.getSolidDataset(
+        testContainerIri,
+        {
+          fetch: resourceOwnerSession.fetch,
+        }
+      );
+
+      const parentContainerContainsAll =
+        solidClient.getThingAll(parentContainer);
```
NSeydoux, line 542:
```suggestion
        solidClient.getUrlAll(parentContainer, LDP.member);
```
We can be more specific in the test here. The LDP namespace may be added in the existing constant files alredy defining a bunch or IRIs to have `http://www.w3.org/ns/ldp#member`. Alternatively, `@inrupt/vocab-common-rdf` may be added as a dependency (but the former should be enough).

src/resource/createContainerInContainer.test.ts
```diff
@@ -0,0 +1,64 @@
...
+  "https://example.come/container/anothercontainer";
+
+describe("createContainerInContainer", () => {
+  it("authenticates using the provided VC", async () => {
+    const solidClientModule = jest.requireMock("@inrupt/solid-client") as any;
+    const mockedDataset = mockSolidDatasetFrom("https://some.url");
+    solidClientModule.createContainerInContainer.mockResolvedValueOnce(
+      MOCKED_DATASET
+    );
+    const mockedFetch = jest.fn() as typeof fetch;
```
NSeydoux, line 46:
```suggestion
    const mockedFetch = jest.fn<typeof fetch>();
```

src/resource/createContainerInContainer.test.ts
```diff
@@ -0,0 +1,64 @@
...
+  return solidClientModule;
+});
+
+const MOCKED_DATASET = mockSolidDatasetFrom("https://some.url");
+const TEST_CONTAINER_URL: UrlString =
+  "https://example.come/container/anothercontainer";
+
+describe("createContainerInContainer", () => {
+  it("authenticates using the provided VC", async () => {
+    const solidClientModule = jest.requireMock("@inrupt/solid-client") as any;
```
NSeydoux, line 41:
```suggestion
    const solidClientModule = jest.requireMock("@inrupt/solid-client") as jest.Mocked<typeof SolidClient>;
```
with the added import `import type * as SolidClient from "@inrupt/solid-client";`. It's a cleaner way to do things I've become familiar with after a lot of this code was written, which is why it's not used everywhere. I try to clean things up when we touch different portions of the codebase :)

src/resource/createContainerInContainer.ts
```diff
@@ -0,0 +1,82 @@
...
+} from "@inrupt/solid-client";
+import { VerifiableCredential } from "@inrupt/solid-client-vc";
+import { fetchWithVc } from "../fetch";
+import { FetchOptions } from "../type/FetchOptions";
+
+/**
+ * Create an empty Container inside the Container at the given URL.
+ *
+ * Throws an error if creating the Container failed, e.g. because the current user does not have
+ * permissions to.
```
NSeydoux, line 36:
```suggestion
 * Throws an error if creating the Container failed, e.g. because the current user does not have
 * permissions to. In particular, the Access Grant being used should at least append access to the target Container.
```

src/resource/createContainerInContainer.ts
```diff
@@ -0,0 +1,82 @@
...
+ * a Container, but is allowed to add new files; in other words, they have Append, but not Write
+ * access to a Container. This is useful in situations where someone wants to allow others to,
+ * for example, send notifications to their Pod, but not to view or delete existing notifications.
+ * You can pass a suggestion for the new Resource's name, but the server may decide to give it
+ * another name — for example, if a Resource with that name already exists inside the given
+ * Container.
+ * If the user does have access to write directly to a given location, [[createContainerAt]]
+ * will do the job just fine, and does not require the parent Container to exist in advance.
+ *
+ * @param containerUrl URL of the Container in which the empty Container is to be created.
```
NSeydoux, line 50: There's an extra parameter compared to `solid-client`, the `accessGrant`, which should be mentioned here.

src/resource/createContainerInContainer.ts
```diff
@@ -0,0 +1,82 @@
...
+ * @param options Optional parameter `options.fetch`: An alternative `fetch` function to make the HTTP request, compatible with the browser-native [fetch API](https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/fetch#parameters). `options.slugSuggestion` accepts a string for your new Container's name.
+ * @returns A promise that resolves to a SolidDataset with ResourceInfo if successful, and that
+ * rejects otherwise.
+ * @since 0.2.0
```
NSeydoux, line 54:
```suggestion
 * @since unreleased
```

src/resource/createContainerInContainer.ts
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+interface SaveInContainerOptions extends FetchOptions {
+  slugSuggestion?: string;
+}
+
```
NSeydoux, line 60: I'm not sure TSDoc will pick up the doc comment if it's not directly above the function.

src/resource/createContainerInContainer.ts
```diff
@@ -0,0 +1,81 @@
...
+ * @since 0.2.0
+ */
+
+type SaveInContainerOptions = Partial<
+  FetchOptions & {
+    slugSuggestion: string;
+  }
+>;
+export async function createContainerInContainer<SolidDataset>(
+  containerUrl: UrlString,
```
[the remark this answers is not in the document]
NSeydoux: That's fine, it's consistent with the rest of this package.

src/resource/createContainerInContainer.test.ts
solid-akb: Since it's optional and tested in coreCreateContainerInContainer it didn't feel necessary but I will add.

src/resource/createContainerInContainer.ts
solid-akb: Eee I still don't like this. It feels like bloat. What is the benefit again? If I am a developer I can look at the source code and explicitly see what options are available to me?

e2e/node/e2e.test.ts
solid-akb: This was breaking when I ran it locally. I think it's because getUrlAll is looking for a thing as the first argument and parentContainer is a Dataset.

src/resource/createContainerInContainer.ts
ThisIsMissEm: it's mostly about providing a layer of isolation, as `options` here is not exactly the options that `coreCreateContainerInContainer` accepts, if we add an option here, what would happen is that it'd be exposed to `coreCreateContainerInContainer` even if that method doesn't accept that option.

e2e/node/e2e.test.ts
NSeydoux: Oh that's right, you'd have to extract the Thing which IRI matches the container first

NSeydoux (approved):

PR #424, opened 2022-11-30
Feat: extend getter functions to support ODRL data model

VirginiaBalseiro:
# Extend getter functions to support ODRL data model

This PR extends getter functions tu support ODRL data model. 

# Checklist

- [ ] New functions/types have been exported in `index.ts`, if applicable.

### DIFF
--- a/src/common/getters.ts
+++ b/src/common/getters.ts
@@ -199,7 +199,9 @@ export function getIssuanceDate(
 export function getExpirationDate(
   vc: AccessGrantGConsent | AccessRequestGConsent | AccessGrantOdrl
 ): Date | undefined {
-  return typeof vc.expirationDate === "string" ? new Date(vc.expirationDate) : undefined;
+  return typeof vc.expirationDate === "string"
+    ? new Date(vc.expirationDate)
+    : undefined;
 }
 
 /**
--- a/src/common/getters.ts
+++ b/src/common/getters.ts
@@ -22,6 +22,7 @@ import { AccessGrantGConsent } from "../gConsent/type/AccessGrant";
 import { AccessRequestGConsent } from "../gConsent/type/AccessRequest";
 import { isAccessGrant as isGConsentAccessGrant } from "../gConsent/guard/isAccessGrant";
 import { AccessModes, resourceAccessToAccessMode } from "../type/AccessModes";
+import { AccessGrantOdrl, isCredentialAccessGrantOdrl } from "../odrl";
 
 /**
  * Get the resources to which an Access Grant/Request applies.
@@ -29,15 +30,20 @@ import { AccessModes, resourceAccessToAccessMode } from "../type/AccessModes";
  * @example
  *
  * ```
- * const resources = getResources(gConsentGrant);
+ * const resources = getResources(accessGrant);
  * ```
  *
  * @param vc The Access Grant/Request
- * @returns The VC IRI
+ * @returns The resources IRIs
  */
 export function getResources(
-  vc: AccessGrantGConsent | AccessRequestGConsent
+  vc: AccessGrantGConsent | AccessRequestGConsent | AccessGrantOdrl
 ): string[] {
+  if (isCredentialAccessGrantOdrl(vc)) {
+    return vc.credentialSubject.permission.map(
+      (permission) => permission.target
+    );
+  }
   if (isGConsentAccessGrant(vc)) {
     return vc.credentialSubject.providedConsent.forPersonalData;
   }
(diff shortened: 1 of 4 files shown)


src/common/getters.ts
```diff
@@ -22,22 +22,28 @@ import { AccessGrantGConsent } from "../gConsent/type/AccessGrant";
...
  * @param vc The Access Grant/Request
- * @returns The VC IRI
+ * @returns The resources IRIs
  */
 export function getResources(
-  vc: AccessGrantGConsent | AccessRequestGConsent
+  vc: AccessGrantGConsent | AccessRequestGConsent | AccessGrantOdrl
 ): string[] {
+  if (isCredentialAccessGrantOdrl(vc)) {
+    return vc.credentialSubject.permission.map(
```
ThisIsMissEm, line 43: Can any of these be nullable? Can we get just a single value instead of an array here?

src/common/getters.ts
```diff
@@ -96,15 +107,21 @@ export function getRequestor(
...
  *
  * @param vc The Access Grant/Request
  * @returns The access modes the grant recipient can exercise.
  */
 export function getAccessModes(
-  vc: AccessGrantGConsent | AccessRequestGConsent
+  vc: AccessGrantGConsent | AccessRequestGConsent | AccessGrantOdrl
 ): AccessModes {
+  if (isCredentialAccessGrantOdrl(vc)) {
+    const modes = vc.credentialSubject.permission
```
ThisIsMissEm, line 120: Ditto here about if it can be nullable or not an array

src/common/getters.ts
```diff
@@ -171,16 +190,16 @@ export function getIssuanceDate(
...
  * @param vc The Access Grant/Request
- * @returns The issuance date
+ * @returns The expiration date
  */
 export function getExpirationDate(
-  vc: AccessGrantGConsent | AccessRequestGConsent
+  vc: AccessGrantGConsent | AccessRequestGConsent | AccessGrantOdrl
 ): Date | undefined {
-  return vc.expirationDate ? new Date(vc.expirationDate) : undefined;
+  return vc.expirationDate ? new Date(vc.expirationDate as string) : undefined;
```
ThisIsMissEm, line 202:
```suggestion
  return typeof vc.expirationDate === "string" ? new Date(vc.expirationDate) : undefined;
```

By explicitly checking the type, you avoid the need for the `as X`

ThisIsMissEm: What should this function do if it fails to parse the string as a Date?

ThisIsMissEm (commented):
A few questions & a minor change, but looking good!

src/common/getters.ts
```diff
@@ -50,20 +56,22 @@ export function getResources(
...
  * @param vc The Access Grant/Request
  * @returns The resource owner WebID
  */
-export function getResourceOwner(vc: AccessGrantGConsent): string;
 export function getResourceOwner(
-  vc: AccessGrantGConsent | AccessRequestGConsent
+  vc: AccessGrantGConsent | AccessGrantOdrl
+): string;
+export function getResourceOwner(
+  vc: AccessGrantGConsent | AccessRequestGConsent | AccessGrantOdrl
```
NSeydoux, line 69: I think I screwed up the overload initially, because I don't think the signature
```
export function getResourceOwner(vc: AccessGrantGConsent): string;
export function getResourceOwner(
  vc: AccessGrantGConsent | AccessRequestGConsent
): string | undefined;
export function getResourceOwner(
  vc: AccessGrantGConsent | AccessRequestGConsent
): string | undefined
```
makes sense, it should rather be 
```
export function getResourceOwner(vc: AccessGrantGConsent): string;
// No longer a type disjunction here
export function getResourceOwner(vc: AccessRequestGConsent): string | undefined;
export function getResourceOwner(
  vc: AccessGrantGConsent | AccessRequestGConsent
): string | undefined
```
and so that means the new signature would look like 
```
export function getResourceOwner(vc: AccessGrantGConsent | AccessGrantOdrl): string;
export function getResourceOwner(vc: AccessRequestGConsent): string | undefined;
export function getResourceOwner(
  vc: AccessGrantGConsent | AccessRequestGConsent | AccessGrantOdrl
): string | undefined
```

src/common/getters.ts
NSeydoux: I'd expect it to throw, that's what the call to `new Date(vc.expirationDate)` should do right?

src/common/getters.ts
ThisIsMissEm: Why would this not return `string | undefined`?
``` 
export function getResourceOwner(vc: AccessGrantGConsent | AccessGrantOdrl): string;
```

src/common/getters.ts
ThisIsMissEm: I'd expect it to return undefined, why would it throw? An invalid date is as good as no date, surely?

ThisIsMissEm: Right but you can't mutate what is passed in, this is an issue with the VC provider, they should've validated & formatted the date according to the JSON spec.

NSeydoux: That's true, this is coming from the server and it is immutable, so throwing doesn't really help actually. I never feel too good about swallowing errors, if something is wrong with the VC it would seem right to bubble it up in some way, and have the consuming app be able to notify the user that something is wrong and they may want to do something about it. From a usefulness perspective, an invalid date is as good as no date, but I guess having a different behaviour allows the app to act accordingly?

src/common/getters.ts
NSeydoux: A Grant being issued by the resource owner, it is mandatory that the VC subject is the resource owner. The resource owner may not be defined in an Access Request, because a Request may be scoped to a target resource with access modes, without specifying who should grant access.

src/common/getters.ts
VirginiaBalseiro: According to the type definition. here: https://github.com/inrupt/solid-client-access-grants-js/blob/7f2d36a31943a1af238a20626acfeab925b64bf6/src/odrl/type/AccessGrant.ts#L52, `permission` should be an array and required.

src/common/getters.ts
VirginiaBalseiro: same as above

src/common/getters.ts
NSeydoux: My bad, I assumed the disjunction would be picked up from the overload, which isn't correct, so @VirginiaBalseiro you were right.

NSeydoux (approved):

PR #447, opened 2022-12-19
Lookup recursive access grant

NSeydoux:
The issue this intends to solve is the following: when recursive access grants are supported, looking up an access grant applicable to a given resource may not include said resource, and instead include one of its ancestor containers. To solve this, when looking up access grants for a target resource, requests for recursive access grants are sent to each ancestor container of the resource. The result set is then filtered to remove non-recursive grants on ancestor resources.

Grants are assumed to be implicitly recursive, so only explicitly non-recursive grants are filtered out. Explicitly non-recursive grants targetting the target resource are not filtered out.

This PR also has some refactor of access grant mocks to make unit tests easier. It also has some changes in the getters and the AccessCredentials type to add the `inherit` flag that may conflict with another PR, so these could be removed when rebasing.

# Checklist

- [ ] Relevant documentation, if any, has been written/updated.

### DIFF
--- a/src/gConsent/manage/getAccessGrantAll.ts
+++ b/src/gConsent/manage/getAccessGrantAll.ts
@@ -57,8 +57,10 @@ const getAncestorUrls = (resourceUrl: URL) => {
 };
 
 /**
- * Retrieve Access Grants issued over a resource.
- * Can be filtered by requestor, access and purpose.
+ * Retrieve Access Grants issued over a resource. The Access Grants may be filtered
+ * by requestor, access modes and purpose. In order to discover all applicable Access
+ * Grants for the target resource, including recursive Access Grants issued over
+ * an ancestor container, the resources hierarchy is walked up to the storage root.
  *
  * @param resource The URL of a resource to which access grants might have been issued.
  * @param grantShape The properties of grants to filter results.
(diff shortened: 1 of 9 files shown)


NSeydoux:
Rebased on https://github.com/inrupt/solid-client-access-grants-js/pull/447, review that first.

chelseapinka (approved):

chelseapinka:
it does seem like you might need a longer timeout though

src/gConsent/manage/getAccessGrantAll.ts
```diff
@@ -32,10 +32,35 @@ import type { IssueAccessRequestParameters } from "../type/IssueAccessRequestPar
...
 import type { BaseGrantBody } from "../type/AccessVerifiableCredential";
+import { isAccessGrant } from "../guard/isAccessGrant";
+import { isBaseAccessGrantVerifiableCredential } from "../guard/isBaseAccessGrantVerifiableCredential";
+import { AccessGrant } from "../type/AccessGrant";
+import { getInherit, getResources } from "../../common/getters";
+
+// Iteratively build the list of ancestor containers from the breakdown of the
+// resource path: for resource https://pod.example/foo/bar/baz, we'll want the result
+// to be ["https://pod.example/", "https://pod.example/foo/", "https://pod.example/foo/bar/", https://pod.example/foo/bar/baz].
+const getAncestorUrls = (resourceUrl: URL) => {
```
ThisIsMissEm, line 43: Shouldn't this be following the containment triples/link headers? Rather than being directory based?

PR #461, opened 2023-01-06
JSON-LD status validation

NSeydoux:
We currently aren't parsing Access Grant as JSON-LD, but only as JSON. This results in some validations being stricter than necessary: any IRI was expected fully qualified, while it is acceptable to get it in its abbreviated form based on the JSON-LD context.


ThisIsMissEm (changes requested):
One API change detected, but otherwise good

src/gConsent/guard/isAccessGrant.ts
```diff
@@ -33,7 +30,7 @@ export function isAccessGrant(
 ): vc is BaseAccessVcBody & AccessGrantBody {
   return (
     (vc as AccessGrantBody).credentialSubject.providedConsent !== undefined &&
-    [GC_CONSENT_STATUS_EXPLICITLY_GIVEN, GC_CONSENT_STATUS_DENIED].includes(
+    CONSENT_STATUS.has(
```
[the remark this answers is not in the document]
NSeydoux: [bc4b013](https://github.com/inrupt/solid-client-access-grants-js/pull/461/commits/bc4b013d30c584fa931df562ec89b6b5e78fe35f) should fix this.

CHANGELOG.md
```diff
@@ -30,6 +30,11 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - The `getters` module contains functions and a class to interact with Verifiable
   Credentials without having to hard-code the data model in your application.
 
+### Bugfixes
+
+- The Access Grants and Access Requests status is now accepted in its abbreviated
+  form, and not only as a fully qualified IRI, as allowed by the JSON-LD context.
```
ThisIsMissEm, line 36: Is it worth specifying here that we did not decide to parse access grants as JSON-LD due to technical infrastructure around JSON-LD parsing? And that we're aware JSON parsing isn't quite correct here.

src/gConsent/guard/isGConsentAttributes.ts
```diff
@@ -34,7 +34,7 @@ function isResourceAccessModeArray(x: unknown): x is Array<ResourceAccessMode> {
 }
 
 function isGConsentStatus(x: unknown): x is GConsentStatus {
-  return typeof x === "string" && (CONSENT_STATUS as Set<string>).has(x);
+  return typeof x === "string" && (ACCESS_STATUS as Set<string>).has(x);
```
ThisIsMissEm, line 37:
```suggestion
  return typeof x === "string" && ACCESS_STATUS.has(x);
```

`ACCESS_STATUS` should already be a `Set<string>` per its definition above?

src/gConsent/manage/approveAccessRequest.test.ts
```diff
@@ -311,6 +312,24 @@ describe("approveAccessRequest", () => {
...
   });
 
+  it("accepts the returned VC using the abbreviated status instead of the fully qualified IRI", async () => {
+    mockAccessApiEndpoint();
+    const mockedVcModule = jest.requireMock(
+      "@inrupt/solid-client-vc"
+    ) as typeof VcClient;
+    const mockedIssue = jest.spyOn(mockedVcModule, "issueVerifiableCredential");
+    const mockedVc = mockAccessGrantVc();
+    mockedVc.credentialSubject.providedConsent.hasStatus =
```
ThisIsMissEm, line 322: Can we also update other places these IRIs are used with the constants? e..g, line 299 in this file.

ThisIsMissEm (commented):
I think this is good, just two small comments

CHANGELOG.md
NSeydoux: Good point, that would be more visible than the internal comment I left about this.

NSeydoux: 4532886d927ea9cf32f69f5c58cbe8f70290d526

src/gConsent/manage/approveAccessRequest.test.ts
NSeydoux: Actually, I will do it the other way around and update this to use the string value: 941bbe788c1dcea355d37a530f3e1ba1ee6a4f24. The reason is that if we use the constants in the tests and in the code, we could update the constants to whatever value and the tests would still pass, whereas having the actual value in the test shows intent of what the constant should be.

src/gConsent/guard/isGConsentAttributes.ts
ThisIsMissEm: Is this just because we're merging two sets? What if we define the set members internally as two arrays, and then export 3 sets? Would that fix?

src/gConsent/manage/approveAccessRequest.test.ts
ThisIsMissEm: Good call.

ThisIsMissEm (changes requested):
I would like for us to resolve this comment before merging: https://github.com/inrupt/solid-client-access-grants-js/pull/461/files#r1067302621

src/gConsent/guard/isGConsentAttributes.ts
NSeydoux: What about https://github.com/inrupt/solid-client-access-grants-js/pull/461/commits/1bfe0f8a1b47824297aea5f332afc0b0d0359216 ? It also makes the sets read-only.

ThisIsMissEm: I like it, but I don't understand it.

NSeydoux: The main difference boils down to
```
const stringArray = ["a", "b", "c"] // inferred to string[]
const narrowArray = ["a", "b", "c"] as const // inferred to  readonly ["a", "b", "c"]
```
The sets being generic, they just reflect the type of the collection they wrap around. The change in https://github.com/inrupt/solid-client-access-grants-js/commit/1bfe0f8a1b47824297aea5f332afc0b0d0359216 moves from the latter to the former.

ThisIsMissEm: Okay, cool!

ThisIsMissEm (approved):
LGTM

PR #477, opened 2023-01-23
Set recursive ACP VC access to match grant recursiveness

NSeydoux:
This enables end-to-end tests for recursive access grants against ESS dev-next, which is the environment where they are supported. This involves a couple of things:
- Upgrading to the latest `@inrupt/solid-client`, which adds the possibility to set a recursive VC matcher updating a resource's ACR
- Updating the end-to-end tests to only run recursive access grants tests against environment supporting them
- The `@inrupt/solid-client` upgrade comes with a couple of unplanned changes, such as `cross-fetch` mock updates, and adding `@types/rdfjs__dataset` as a dependency to fix docs generation.


e2e/node/e2e.test.ts
```diff
@@ -1014,52 +1016,65 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
...
-        fetch: requestorSession.fetch,
-      });
-
-      await expect(requestorFile.text()).resolves.toBe(testFileContent);
+    const testIf = (condition: boolean) => (condition ? it : it.skip);
+    // Conditional tests confuse eslint.
+    /* eslint-disable jest/no-standalone-expect */
+
+    // Only enable this test in environments supporting recursive access grants
+    testIf(environmentFeatures?.RECURSIVE_ACCESS_GRANTS === "true")(
```
ThisIsMissEm, line 1024: What if we add it to our testing framework? `const featureTest = createFeatureTests("RECURSIVE_ACCESS_GRANTS", jest)`

src/gConsent/discover/redirectToAccessManagementUi.browser.test.ts
```diff
@@ -114,8 +115,9 @@ describe("redirectToAccessManagementUi", () => {
         }
       );
       // Yield the event loop to make sure the blocking promises completes.
+      // FIXME: Why is setImmediate undefined in this context ?
```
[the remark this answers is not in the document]
ThisIsMissEm: But really this code is a bit weird as it's an "await for 1 tick of the event loop then grab `window.location`"

what would be better would be to spy on `window.location` so you can observe it's changes

src/gConsent/manage/approveAccessRequest.ts
```diff
@@ -108,7 +116,9 @@ async function addVcMatcher(
         );
       }
       // eslint-disable-next-line camelcase
-      const updatedResource = acp_ess_2.setVcAccess(resourceInfo, accessMode);
+      const updatedResource = acp_ess_2.setVcAccess(resourceInfo, accessMode, {
+        inherit: options?.inherit ?? true,
```
ThisIsMissEm, line 120: We're defaulting this to true now?

src/gConsent/manage/getAccessGrant.test.ts
```diff
@@ -20,17 +20,30 @@
...
 
-jest.mock("cross-fetch");
+jest.mock("cross-fetch", () => {
+  const crossFetch = jest.requireActual("cross-fetch") as jest.Mocked<
+    typeof CrossFetch
+  >;
+  return {
+    // Do no mock the globals such as Response.
+    ...crossFetch,
+    Response: crossFetch.Response,
```
ThisIsMissEm, line 37:
```suggestion
```

this is handled by the line above, or should be.

src/gConsent/request/issueAccessRequest.test.ts
```diff
@@ -460,6 +471,7 @@ describe("issueAccessRequest", () => {
             normalizedAccessRequest.credentialSubject.hasConsent
               .forPersonalData[0],
           mode: normalizedAccessRequest.credentialSubject.hasConsent.mode[0],
+          inherit: "true",
```
ThisIsMissEm, line 474: Do we need a test case for inherit === false?

e2e/node/e2e.test.ts
```diff
@@ -944,9 +948,7 @@ describe(`End-to-end access grant tests for environment [${environment}}]`, () =
     });
   });
 
-  // This is skipped until ESS supports recursive Access Grants
-  // eslint-disable-next-line jest/no-disabled-tests
-  describe.skip("access grants can be recursive or only apply to the target resource", () => {
+  describe("access grants can be recursive or only apply to the target resource", () => {
```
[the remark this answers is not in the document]
ThisIsMissEm: Note on the two disabled tests above, those can probably use inherit / recursive access grants to function as expected

ThisIsMissEm (approved):
A few small comments, but LGTMO.

package.json
```diff
@@ -127,8 +127,9 @@
     "typescript": "^4.8.4"
   },
   "dependencies": {
-    "@inrupt/solid-client": "^1.23.3",
+    "@inrupt/solid-client": "^1.25.0",
     "@inrupt/solid-client-vc": "^0.5.0",
+    "@types/rdfjs__dataset": "^1.0.5",
```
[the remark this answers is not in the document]
NSeydoux: Absolutely

src/gConsent/manage/approveAccessRequest.ts
NSeydoux: Indeed: the Access Grants are defaulting to recursive, so the ACR structure should do the same otherwise recursive Access Grants aren't applicable.

ThisIsMissEm: Is that how product is rolling this out? Shouldn't they default to non-recursive in this major release (as was previously the case), and then on the next semver major we make them recursive by default?

e2e/node/e2e.test.ts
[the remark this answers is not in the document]
NSeydoux: I rephrased the describe, because I think it makes sense as a single block, and that's a good point, I have re-enabled them: 34f00a0d742b846075b69b360b5b6d25dd14a3ea

src/gConsent/manage/getAccessGrant.test.ts
NSeydoux: Oops yeah looks like a leftover of a test 03265e0161d371884234c0951c685076bb738fc2

src/gConsent/request/issueAccessRequest.test.ts
NSeydoux: Never too sure: d6ad1c5fc7857e295c3292f8423a78af21fc8230

e2e/node/e2e.test.ts
NSeydoux: That's a very good idea

src/gConsent/manage/approveAccessRequest.ts
NSeydoux: Everyone we talked to tripped over the current behavior, expecting granting access to a container to also apply on contained resources. So technically yes, but here it's more aligning with expectations.

PR #485, opened 2023-01-26
Fix end-to-end test environment feature flag

NSeydoux:
This PR fixes the end-to-end tests, which were previously skipping the recursive access grant tests in all environments, and were ignoring the feature flags passed through environment variable to control whether the tests should be skipped or not.
The issue was coming from a bug in the environment test helper, which is now fixed.

This also moves the dataset and non-RDF resource overwrite to the recursive access grants test section, as these features require using access grants.


solid-akb (commented):
Looks good to me!

ThisIsMissEm (approved):
LGTM.

chelseapinka (approved):

PR #504, opened 2023-02-10
Investigate end-to-end flakyness

NSeydoux:
A couple of improvements to improve the reliability of tests, as well as their debuggability: 
- Retry on failed tests
- Do not use static resource names, so that there is no collision on retry
- Make sure that failure of the test doesn't prevent the cleanup, by catching any errors that may happen before deleting the test resource and container
- Add a distinctive user agent to all the HTTP interaction to help going through the logs.


solid-akb (approved):
I appreciate what you did with this PR so much. It was a vague ticket with a bit of an open end but these fixes all look like they will help us. 🙌

Excited to see this implemented!

PR #509, opened 2023-02-15
Remove slug usage from end-to-end tests

NSeydoux:
Some servers will use slugs as a definitive name for the resource, rather than a name that should appear in the created resource IRI. This may result in some cases in a 409 conflict response, if the previous run wasn't cleaned up properly.


solid-akb (approved):
This is an awesome PR. Tough with such an open ended ticket but you crushed it. 💪

PR #579, opened 2023-05-10
chore/deps solid client vc 0.6.0

jeswr:
- Chore(deps): Bump @inrupt/solid-client-vc from 0.5.0 to 0.6.0
- fix: resolve test type errors

Supercedes https://github.com/inrupt/solid-client-access-grants-js/pull/576

### DIFF
--- a/e2e/node/e2e.test.ts
+++ b/e2e/node/e2e.test.ts
@@ -82,8 +82,8 @@ const TEST_USER_AGENT = `Node-based solid-client-access-grant end-to-end tests r
   process.env.CI === "true" ? "in CI" : "locally"
 }`;
 const addUserAgent =
-  (myFetch: typeof fetch, agent: string): typeof fetch =>
-  (input: Parameters<typeof fetch>[0], init: Parameters<typeof fetch>[1]) =>
+  (myFetch: typeof fetch, agent: string) =>
+  (input: RequestInfo | URL, init?: RequestInit | undefined) =>
     myFetch(input, {
       ...init,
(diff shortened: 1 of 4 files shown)


e2e/node/e2e.test.ts
```diff
@@ -82,8 +82,8 @@ const TEST_USER_AGENT = `Node-based solid-client-access-grant end-to-end tests r
   process.env.CI === "true" ? "in CI" : "locally"
 }`;
 const addUserAgent =
-  (myFetch: typeof fetch, agent: string): typeof fetch =>
-  (input: Parameters<typeof fetch>[0], init: Parameters<typeof fetch>[1]) =>
```
NSeydoux, line 86: Why change the utility types into a duplicate of the underlying types?

NSeydoux (approved):

NSeydoux:
https://github.com/inrupt/solid-client-access-grants-js/pull/579/commits/cf21c1c15d45d7e4f1de472f242fa9e6bfd6a2c1 cleans up the jest mocking to use more up-to-date typing, the way mocks were typed is legacy and I thought had been updated when we moved to universal-fetch. Usually, type assertions are a red flag, and should be avoided as much as possible unless there is a good reason to have them.

e2e/node/e2e.test.ts
jeswr: There was a type error because the underlying type was overloaded with one of the overloads not having a second parameter.

So the type error was the Parameter<type of fetch>[1] possibly indexing a const array of length 1.

PR #550, opened 2023-04-05
chore: deprecate buffer api

jeswr:

PR #602, opened 2023-05-19
chore: deprecate buffer API

jeswr:
Supercedes #550.

For those reviewing - note that the main diff you are interested in is [here](https://github.com/inrupt/solid-client-access-grants-js/pull/602/commits/ae59c08b24ff9a497edbb176d2e703cb3cb97886).

Most of the line changes were introduced in [here](https://github.com/inrupt/solid-client-access-grants-js/pull/602/commits/2ad915aeaa048b674cf3e7986a777830a585671c) - which added [this `describe.each`](https://github.com/inrupt/solid-client-access-grants-js/blob/2ad915aeaa048b674cf3e7986a777830a585671c/e2e/node/e2e.test.ts#L114-L129) behavior to test intialisation with both `Buffer` and `File`s.

### DIFF
--- a/src/resource/file.ts
+++ b/src/resource/file.ts
@@ -21,11 +21,13 @@
 
 import {
   UrlString,
+  WithResourceInfo,
   getFile as coreGetFile,
   overwriteFile as coreOverwriteFile,
   saveFileInContainer as coreSaveFileInContainer,
 } from "@inrupt/solid-client";
 import { VerifiableCredential } from "@inrupt/solid-client-vc";
+import type { Buffer as NodeBuffer, File as NodeFile } from "buffer";
 import { fetchWithVc } from "../fetch";
 import { FetchOptions } from "../type/FetchOptions";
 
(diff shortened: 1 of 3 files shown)


src/resource/file.ts
```diff
@@ -86,12 +88,27 @@ export async function getFile(
...
+/**
+ * @deprecated `overwriteFile` should only have `File` input
+ */
+export async function overwriteFile<T extends File | NodeFile | NodeBuffer>(
+  resourceUrl: UrlString,
+  file: T,
+  accessGrant: VerifiableCredential,
+  options?: FetchOptions & { contentType?: string }
+): Promise<T & WithResourceInfo>;
+export async function overwriteFile<T extends File | NodeFile | NodeBuffer>(
```
NSeydoux, line 106:
```suggestion
export async function overwriteFile<T extends NodeBuffer>(
```
The deprecation notice only needs to apply to the Buffer overload.

src/resource/file.ts
```diff
@@ -130,12 +147,31 @@ export async function overwriteFile(
...
+  file: T,
   accessGrant: VerifiableCredential,
   options?: FetchOptions & { contentType?: string; slug?: string }
-) {
+): Promise<T & WithResourceInfo>;
+/**
+ * @deprecated `saveFileInContainer` should only have `File` input; not `Buffer`
+ */
+export async function saveFileInContainer<
+  T extends File | NodeFile | NodeBuffer
```
NSeydoux, line 160:
```suggestion
  T extends NodeBuffer
```
ditto

src/resource/file.ts
jeswr: This set ensures that the deprecation notice applies if I provide something of type `Buffer | File`.

The order of evaluation of the type overloads means that if supply something of *just* `File` or just `NodeFile` then it will be caught by the first overload and the depreciation notice will not apply.

src/resource/file.ts
jeswr: See above comment

src/resource/file.ts
NSeydoux: After a discussion, I realized my proposed change would actually be a breaking change because no overload would then match the legacy signature (just writing this down for docs puprose).

NSeydoux (approved):

PR #672, opened 2023-07-18
Update access grants demo dependencies

kay-kim:
Updated the dependencies for the demo app.

- Also, for Readme.md, we don't have a build:local in our packages.json so the instruction seems confusing.

### DIFF
--- a/examples/solid-client-access-grants-demo/package.json
+++ b/examples/solid-client-access-grants-demo/package.json
@@ -17,8 +17,8 @@
     "start": "node ."
   },
   "dependencies": {
-    "@inrupt/solid-client-access-grants": "file:../../",
-    "@inrupt/solid-client-authn-node": "^1.12.1",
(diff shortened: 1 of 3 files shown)


examples/solid-client-access-grants-demo/package.json
```diff
@@ -17,8 +17,8 @@
     "start": "node ."
   },
   "dependencies": {
-    "@inrupt/solid-client-access-grants": "file:../../",
-    "@inrupt/solid-client-authn-node": "^1.12.1",
+    "@inrupt/solid-client-access-grants": "^2.3.2",
```
chelseapinka, line 20: I dont think we should be targetting a released version of this lib since part of the usefulness of this demo app is for developers to use it while developing new features/fixes and running it with the demo to test out new things.

chelseapinka (changes requested):

kay-kim:
This was just a temp work-around for Monday night. There is a proper JIRA ticket for this.

PR #711, opened 2023-08-16
Fix test app redirect URL

NSeydoux:
The redirect URL was grabbed from window.location, which resulted in it
including OpenID query params.

This is a red herring that I followed trying to fix something else.

### DIFF
--- a/e2e/browser/test-app/components/accessGrants/index.tsx
+++ b/e2e/browser/test-app/components/accessGrants/index.tsx
@@ -157,14 +157,14 @@ export default function AccessGrant({
       <div>
         <button
           type="button"
-          onClick={ handleCreate }
+          onClick={handleCreate}
           data-testid="create-resource"
         >
           Create resource
         </button>
         <button
           type="button"
-          onClick={ handleDelete }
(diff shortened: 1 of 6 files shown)


chelseapinka (approved):

e2e/browser/test-app/components/accessGrants/index.tsx
```diff
@@ -80,7 +81,7 @@ export default function AccessGrant({
     setSharedResourceIri(getSourceUrl(savedFile));
   };
 
-  const handleDelete = async (e) => {
+  const handleDelete = async (e: any) => {
```
jeswr, line 84: Cam we use a more specific type than any?

NSeydoux: I think we could, but actually we don't need the event to go through at all if we just properly use HTML5 (as I learnt writing this code ^^): https://github.com/inrupt/solid-client-access-grants-js/pull/711/commits/d1d5a568a847da55c20ea0ed2e1299b390ad0c68

jeswr (approved):

PR #733, opened 2023-09-04
chore: improve e2e testing

jeswr:
- chore: run e2e tests
- chore: use URL constructors rather than string literals for /verify endpoint
- chore: add 1hr expiries to access requests


NSeydoux (approved):

NSeydoux:
You mention using the URL constructor in the PR description, but I don't see any related change in the code change, am I missing anything?

PR #741, opened 2023-09-10
feat: allow query of denied access grants

jeswr:
Support for querying denied access grants.

In this PR I have also:
 - Fixed e2e test failure noise by:
   - logging in once and keeping the session refreshed, rather than logging in for every test
   - Removing unecessary repetition in tests
   - adding expiry dates to access requests to avoid old ones clogging up filter tests
   - using `retryAsync` to make individual calls more reliable
The e2e tests now run about 4-8x faster and are consistently passing.
 - Skipped tests related to expiry dates of denied requests on dev-2-2

# New feature description

# Checklist

- [x] New modules (i.e. new `.ts` files) are listed in the `exports` field in `package.json`, if applicable.
- [x] New modules (i.e. new `.ts` files) are listed in the `typedocOptions.entryPoints` field in `tsconfig.json`, if applicable.


src/gConsent/constants.ts
```diff
@@ -98,6 +98,7 @@ export const ACCESS_CREDENTIAL_TYPE = new Set([
   CREDENTIAL_TYPE_ACCESS_REQUEST,
   CREDENTIAL_TYPE_ACCESS_GRANT,
   CREDENTIAL_TYPE_ACCESS_DENIAL,
+  "vc:SolidAccessDenial",
```
jeswr, line 101: Denied grants currently use prefixes. So the client side filtering needs to check for this format.

src/gConsent/manage/getAccessGrantAll.ts
```diff
@@ -38,17 +40,21 @@ import { getSessionFetch } from "../../common/util/getSessionFetch";
...
 import { isBaseAccessGrantVerifiableCredential } from "../guard/isBaseAccessGrantVerifiableCredential";
-import type { AccessGrant } from "../type/AccessGrant";
 import { getInherit, getResources } from "../../common/getters";
 import { normalizeAccessGrant } from "./approveAccessRequest";
 
 export type AccessParameters = Partial<
   Pick<IssueAccessRequestParameters, "access" | "purpose"> & {
     requestor: string;
     resource: URL | UrlString;
+    status?: ("granted" | "denied")[];
```
NSeydoux, line 50: Is it necessary to make this an array? Since there are only two options possible, and this is for filtering, you either want one of these values ("only the granted ones") or you don't filter on it and don't provide a value at all. 

This is an issue with backwards compatibility though, since we want the current behavior to remain unchanged. I think we can work around this by introducing a third shorthand, `all`, which would behave the same way as an `undefined` value in the eventual API.

src/gConsent/manage/getAccessGrantAll.ts
```diff
@@ -38,17 +40,21 @@ import { getSessionFetch } from "../../common/util/getSessionFetch";
...
 import { isBaseAccessGrantVerifiableCredential } from "../guard/isBaseAccessGrantVerifiableCredential";
-import type { AccessGrant } from "../type/AccessGrant";
 import { getInherit, getResources } from "../../common/getters";
 import { normalizeAccessGrant } from "./approveAccessRequest";
 
 export type AccessParameters = Partial<
   Pick<IssueAccessRequestParameters, "access" | "purpose"> & {
     requestor: string;
     resource: URL | UrlString;
+    status?: "granted" | "denied" | "all";
```
NSeydoux, line 50: Can you add a comment summarizing the discussion we had, explaining how the `all` value is necessary for backwards compatibility?

src/gConsent/manage/getAccessGrantAll.ts
```diff
@@ -97,20 +103,39 @@ async function internal_getAccessGrantAll(
...
-        providedConsent: {
-          hasStatus: GC_CONSENT_STATUS_EXPLICITLY_GIVEN,
-          forPersonalData: url ? [url.href] : undefined,
-          forPurpose: params.purpose,
-          isProvidedTo: params.requestor,
-          mode: specifiedModes.length > 0 ? specifiedModes : undefined,
-        },
-      },
-    }));
+    statusFull.flatMap(
```
NSeydoux, line 118: Instead of querying for both denied and approved grants if `all` is specified, I would have expected we just don't filter at all on the status. This would result in not having a value in the VC shape for the status, instead of sending two queries, one for each value.

Eventually, we'll want this to be the bahavior on `undefined`, but for backwards compatibility `undefined` maps to approved. So the possible cases are: 
- approved or undefined => status is set in the shape to approved
- denied => status is set to denied
- all => status isn't set, and all values for the status will be accepted on the Holder side.

Does this make sense, or am I missing something?

e2e/node/e2e.test.ts
```diff
@@ -111,99 +143,459 @@ const SHARED_FILE_CONTENT = "Some content.\n";
...
+    const savedFile = await retryAsync(() =>
+      sc.saveFileInContainer(
+        resourceOwnerPod,
+        Buffer.from(SHARED_FILE_CONTENT),
+        {
+          fetch: addUserAgent(
+            addUserAgent(resourceOwnerSession.fetch, TEST_USER_AGENT),
+            TEST_USER_AGENT,
+          ),
+          slug: v4(),
```
NSeydoux, line 171: Instead of putting a random slug, I would just let the server assign a resource name itself. That's functionally similar, but show a different intention if the client doesn't require a specific slug.

EDIT: I see this change comes from a dedicated commit. Is there a specific reason to have this name arbitrarily picked client-side?

NSeydoux (changes requested):

e2e/node/e2e.test.ts
jeswr: I was seeing more VCs than expected returned in https://github.com/inrupt/solid-client-access-grants-js/actions/runs/6197634526/job/16826563949?pr=741#step:6:272.

I thought it may have been because the server was re-using a name. This was not the reason. Still investigating what is happening but this commit can be reverted.

src/gConsent/manage/getAccessGrantAll.ts
jeswr: I've tried that over [here](https://github.com/inrupt/solid-client-access-grants-js/pull/756) and it seems to work.

My main concern was that there would be noise from other things like access requests because we are both setting `hasStatus` but we are also being less restrictive in the `type` field.

No extra noise appears to be turning in those e2e tests so I think it is fine to merge that PR into this one and proceed if you're understanding of the VC service implementation supports this not being an issue.

NSeydoux: There are other differences between Access grants and Access Requests beyond the `status`, the predicates used in the `credentialSubject` will differ, with `providedConsent` in Grants and `hasConsent` (or something similar, this is from top of head) in Requests. All this to say, ever wihtout the status being set, no Access Request will match this request, so I think we should be good there

NSeydoux (approved):

PR #884, opened 2023-12-21
Node16 EoL: remove universal fetch

NSeydoux:
Drop usage of `@inrupt/universal-fetch`, replacing it with the native fetch, globally available in all environments starting with Node 18.

### DIFF
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -20,3 +20,6 @@
 //
 
 import "@inrupt/jest-jsdom-polyfills";
+globalThis.fetch = () => {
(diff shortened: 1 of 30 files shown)


jest.setup.ts
```diff
@@ -20,3 +20,6 @@
 //
 
 import "@inrupt/jest-jsdom-polyfills";
+globalThis.fetch = () => {
+    throw new Error('Global fetch should not be called without being mocked in unit tests')
+}
```
NSeydoux, line 25: Good idea

jeswr (approved):

PR #896, opened 2024-01-10
Fix filtering of non-recursive grants

NSeydoux:
When calling `getAccessGrantAll`, if not filtering grants by target resource, non-recursive grants should not be excluded of the result set.

### DIFF
--- a/src/gConsent/manage/getAccessGrantAll.test.ts
+++ b/src/gConsent/manage/getAccessGrantAll.test.ts
@@ -436,6 +436,31 @@ describe("getAccessGrantAll", () => {
       getAccessGrantAll({ resource: resource.href }),
     ).resolves.toStrictEqual([mockedGrant]);
   });
+
+  it("accepts explicitly non-recursive grants if the target resource is left undefined", async () => {
+    const mockedGrant = await mockAccessGrantVc({
+      inherit: false,
+      resources: [resource.href],
+    });
+    (
+      VcModule as jest.Mocked<typeof VcLibrary>
+    ).getVerifiableCredentialAllFromShape
+      .mockResolvedValueOnce([mockedGrant])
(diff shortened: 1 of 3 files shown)


src/gConsent/manage/getAccessGrantAll.ts
```diff
@@ -249,15 +249,13 @@ async function getAccessGrantAll(
         (vc) => isBaseAccessGrantVerifiableCredential(vc) && isAccessGrant(vc),
       );
   }
-
-  return result.filter(
-    (vc) =>
-      // Explicitly non-recursive grants are filtered out, except if they apply
-      // directly to the target resource.
+  return result.filter((vc) => {
+    return (
```
chelseapinka, line 253: Its interesting that the comment you removed here seems to imply the incorrect behaviour was part of the original design approach.

src/gConsent/manage/getAccessGrantAll.test.ts
```diff
@@ -436,6 +436,31 @@ describe("getAccessGrantAll", () => {
       getAccessGrantAll({ resource: resource.href }),
     ).resolves.toStrictEqual([mockedGrant]);
   });
+
+  it("accepts explicitly non-recursive grants if the target resource is left undefined", async () => {
```
chelseapinka, line 440: :+1:

chelseapinka (approved):
Looks good.

src/gConsent/manage/getAccessGrantAll.ts
NSeydoux: Actually I'll unerase teh coment, but updating it to make it more accurate. It's correct behavior to filter out non-recursive grants if they apply only to an ancestor of the target resource, but that only makes sense if there *is* a target resource ^^

PR #920, opened 2024-02-08
Move browser-based tests away from Podbrowser

NSeydoux:
Browser-based tests no longer redirect to PodBrowser, as it is being deprecated. In addition, this adds exports for request query parameter names. Exporting these constants makes it easier for applications implementing the Access Grant redirection flow to align on parameter names.

Note: #921 will need to be merged first.


jest.config.ts
```diff
@@ -38,10 +38,10 @@ export default {
       // The auto-pickup of fetch is no longer tested because of issues between
       // jose (dependency of solid-client-authn) and ts-jest.
       // FIXME: some unit tests do not cover node-specific code.
-      branches: 95,
-      functions: 95,
-      lines: 95,
-      statements: 95,
+      branches: 94,
```
chelseapinka, line 41: did it end up reducing coverage to add AMC?

chelseapinka (approved):
LGTM once the testing issue is resolved.

jest.config.ts
NSeydoux: It's more that adding AMC uncovered a couple of issues (e.g. the query parameter thing fixed in https://github.com/inrupt/authorization-management-component/pull/150), but doing so ended up in exporting some additional things, thus resulting in a small dip in the existing code coverage (which is based on the public API if I'm correct). Because we were already borderline, it went below the threshold, hence https://github.com/inrupt/solid-client-access-grants-js/pull/921 (which shows up in this PR) and the manual lowering of the threshold to prevent this PR from dragging on. I have created a tech debt ticket to improve the code coverage again.

edwardsph (approved):
LGTM

PR #923, opened 2024-02-09
Support alternative predicates to gc:isProvidedTo

NSeydoux:
The Access Grant data model allows three alternate predicates in their shape to link an Access Grant to the Agent being granted Access:
- gc:isProvidedTo
- gc:isProvidedToPerson
- gc:isProvidedToController

The latter two are now supported by the library.


edwardsph (approved):
LGTM

PR #962, opened 2024-03-22
issueAccessVc checks size of the resources array before grabbing a va…

garciafdezpatricia:
`issueAccessVc` checks now the size of the resources array before grabbing a value from it, consequently showing a comprehensive error when no resources are linked to the access grant or request. 

Also, a comprehensive error is shown when no `accessEndpoint` is provided and hence its discovery fails.

This PR fixes #SDK-3041.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ The following changes are pending, and will be applied on the next major release
 ### Bugfixes
 
 - Removed base64url dependency due to potential issues with the browser environment.
+- Descriptive error shown when trying to create an access request/approve an access grant for no resources.
+- Comprehensive error shown when the issuer endpoint for an access grant/request can't be computed.
 
 ## [3.0.4](https://github.com/inrupt/solid-client-access-grants-js/releases/tag/v3.0.4) - 2024-02-12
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ The following changes are pending, and will be applied on the next major release
 ### Bugfixes
 
 - Removed base64url dependency due to potential issues with the browser environment.
+- Descriptive error shown when trying to create an access request/approve an access grant for no resources.
```
NSeydoux, line 16:
```suggestion
- A descriptive error is now thrown when trying to create an access request/approve an access grant for no resources. Note that previously, such call may not have thrown, but it resulted in an Access Grant granting access to nothing, which would have caused confusion when trying to use it.
```

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ The following changes are pending, and will be applied on the next major release
...
+- Comprehensive error shown when the issuer endpoint for an access grant/request can't be computed.
```
NSeydoux, line 17:
```suggestion
- The error thrown when the issuer endpoint for an access grant/request cannot be computed is more comprehensive.
```

NSeydoux:
Note that the browser-based tests appear to be broken unrelated to your PR: it is also happening in https://github.com/inrupt/solid-client-access-grants-js/pull/957, which doesn't contain any code change.

NSeydoux (approved):

PR #999, opened 2024-04-24
cancelAccessRequest supports revoking an Access Request

garciafdezpatricia:
This PR fixes cancelAccessRequest type checking Access Grants.


src/gConsent/manage/revokeAccessGrant.ts
```diff
@@ -48,19 +49,18 @@ const { quad, namedNode } = DataFactory;
 async function revokeAccessGrant(
   vc: DatasetWithId | VerifiableCredential | URL | UrlString,
   options: Omit<AccessBaseOptions, "accessEndpoint"> = {},
+  type: NamedNode<string> = solidVc.SolidAccessGrant,
```
NSeydoux, line 52: That looks like a good direction. Ideally however, I'd prefer this not to be part of the public API. Could you create a function `revokeAccessCredential` (it can be side-by-side with the current `revokeAccessGrant`) that would have the additional parameter you just added in its signature, and then `revokeAccessGrant` and `cancelAccessRequest` would both call `revokeAccessCredential` with the appropriate parameters.

`revokeAccessCredential` would not be exported by index files that are documented in `package.json`, making it not public API. You can even add the `@internal` TSDoc annotation to make sure the function isn't documented in API docs.

src/gConsent/manage/revokeAccessGrant.ts
```diff
@@ -38,29 +39,23 @@ import { getBaseAccess } from "../util/getBaseAccessVerifiableCredential";
...
-      quad(namedNode(getId(credential)), TYPE, solidVc.SolidAccessGrant),
-    ) &&
+    !credential.has(quad(namedNode(getId(credential)), TYPE, type)) &&
     !credential.has(
       quad(namedNode(getId(credential)), TYPE, solidVc.SolidAccessDenial),
     )
   ) {
     throw new Error(
-      `An error occurred when type checking the VC: Not of type [${solidVc.SolidAccessGrant.value}] or [${solidVc.SolidAccessDenial.value}].`,
+      `An error occurred when type checking the VC: Not of type [${type.value}] or [${solidVc.SolidAccessDenial.value}].`,
```
NSeydoux, line 58:
```suggestion
      `An error occurred when type checking the VC: Not of type [${type.value}].`,
```
An Access Denial would not be accepted in all cases, so this error message should be specific to the provided VC type. However, this means that this is a breaking change: `revokeAccessGrant` supported both approved or denied Access Grants before this change, and now only approved ones.

You could slightly change the internal function so that it accepts a list of types, therefore potentially accepting more than one value.

src/gConsent/util/access.mock.ts
```diff
@@ -37,6 +37,7 @@ import type { AccessGrant } from "../type/AccessGrant";
 import type { AccessRequest } from "../type/AccessRequest";
 
 type RequestVcOptions = Partial<{
+  issuer: string;
```
NSeydoux, line 40: That's good, but you'll want this to have an effect on the resulting VC when calling `mockAccessRequestVcObject`. See how the VC object takes the provided options, or defaults to some arbitrary values if no option is passed in.

NSeydoux (dismissed):

src/gConsent/manage/revokeAccessGrant.ts
```diff
@@ -38,29 +39,27 @@ import { getBaseAccess } from "../util/getBaseAccessVerifiableCredential";
...
- * @param options Optional properties to customise the request behaviour.
- * @returns A void promise.
- * @since 0.4.0
+ * @internal
  */
-async function revokeAccessGrant(
+async function revokeAccessCredential(
   vc: DatasetWithId | VerifiableCredential | URL | UrlString,
   options: Omit<AccessBaseOptions, "accessEndpoint"> = {},
+  type: NamedNode<string>[] = [
```
NSeydoux, line 47:
```suggestion
  types: NamedNode<string>[] = [
```
Arrays are typically named plural as opposed single values.

src/gConsent/manage/revokeAccessGrant.ts
```diff
@@ -38,29 +39,27 @@ import { getBaseAccess } from "../util/getBaseAccessVerifiableCredential";
...
+      (vcType) =>
+        !credential.has(quad(namedNode(getId(credential)), TYPE, vcType)),
     )
   ) {
     throw new Error(
-      `An error occurred when type checking the VC: Not of type [${solidVc.SolidAccessGrant.value}] or [${solidVc.SolidAccessDenial.value}].`,
+      `An error occurred when type checking the VC: Not of type [${type[0].value}] 
+      ${type.length > 1 ? `or [${type[1].value}]` : ``}.`,
     );
   }
```
NSeydoux, line 64: `getBaseAccess` already checks for the type, so it is actually not be necessary at all to perform the check again (which was already present before your PR, it's nothing you've introduced no worries).

Provided the underlying `getBaseAccess` checks for the type and throws on mismatch, I would be tempted to do something like: 
```
  const credential = (
    await Promise.all(
      types.map(async (type) => {
        try {
          return await getBaseAccess(vc, options, type);
        } catch (e) {
          return undefined;
        }
      }),
    )
  ).find((cred) => cred !== undefined);
```
and then test if credential is undefined. If so, the VC was not of the target types, and you can throw an error as previously, and otherwise you can carry on with the rest of the function. Does this make sense?

NSeydoux: Indeed, although you'll want to do something more generic rather than expecting one or two types: 
```
`An error occurred when type checking the VC: Not of the expected types ${types.map((type) => type.value)}`
```
and then line 62 becomes unnecessary.

garciafdezpatricia: There's an "issue" with the new code for computing the credential: there are some specific tests that are expecting the Error thrown by `getBaseAccess`. Because of the `try/catch` this error is handled, and hence the message from the error thrown is not the expected `401 Unauthorized` but the one introduced in line 61.

Should I go ahead and adapt the tests or should we go back to the previous version?

acoburn (approved):

PR #1027, opened 2024-05-30
Defensive statements before looking into VCs

garciafdezpatricia:
Code will now check if the VC provided is of the supported types, namely: DatasetWithId, VerifiableCredential, UrlString or URL. If it is not, it will be parsed into a valid VC before continuing with the execution flow, avoiding the code from breaking (due to the switch from JSON to JSON-LD VCs breaking change introduced in v3).


src/common/verify/isValidAccessGrant.ts
```diff
@@ -28,8 +28,10 @@ import {
   getIssuer,
   getVerifiableCredentialApiConfiguration,
 } from "@inrupt/solid-client-vc";
+import { isUrl } from "@inrupt/solid-client-vc/dist/common/common";
```
NSeydoux, line 31:
```suggestion
import { isUrl } from "@inrupt/solid-client-vc/common";
```
This should be aligned with the `exports` in https://github.com/inrupt/solid-client-vc-js/blob/c18412db41f03aeab1024fef0b63b28bf7ba0bad/package.json#L34.

NSeydoux: Ah interesting, it doesn't resolve the issue.

NSeydoux:
`isUrl` is probably something we'll prefer to locally re-implement (internally, and not exporting it in the public API) insteadu of importing it from the VC library.

NSeydoux:
It looks rollup is also complaining about the subpath import, so there must be something wrong about it, my bad

NSeydoux:
Our current rollup config is set up in a way that prevents this, but I think it would not be too problematic to change. I'll give it a go tomorrow.

src/common/verify/isValidAccessGrant.ts
```diff
@@ -48,7 +51,20 @@ async function isValidAccessGrant(
...
+    typeof vc !== "string" &&
+    !isUrl(vc.toString()) &&
+    !(vc instanceof URL) &&
+    !isDatasetCore(vc)
+  ) {
+    validVC = await verifiableCredentialToDataset(vc, {
+      includeVcProperties: true,
+      requireId: false,
+    });
+  }
```
NSeydoux, line 65:
```suggestion
  if (
    typeof vc === "object" &&
    !isDatasetCore(vc) &&
    typeof (vc as { id: string }).id === "string"    
  ) {
    validVC = await verifiableCredentialToDataset(vc, {
      includeVcProperties: true,
      requireId: true,
    });
  }
```
I think I would prefer the logic here shifted to check whether the object that got passed is indeed a plain JSON representation of a Verifiable Credential, which would have a `.id` property. It is more restrictive than going by elimination, and assuming that a non-string non-URL non-Datasetcore object would be parseable as JSON-LD.

src/common/verify/isValidAccessGrant.ts
```diff
@@ -48,7 +51,20 @@ async function isValidAccessGrant(
...
+  // validVc is null if the provided VC is of the supported types
+  const vcObject = await getBaseAccess(validVC ?? vc, options);
```
NSeydoux, line 67:
```suggestion
  if(validVc === null && isUrl(vc.toString())) {
    validVc = vc;
  }
  if(validVc === null) {
    throw new Error(`Invalid argument: expected either a VC URL or a RDFJS DatasetCore, received ${vc}`)
  }
  
  const vcObject = await getBaseAccess(validVc, options);
```
This allows to add an explicit error being thrown in case we didn't know how to handle the provided argument.

src/common/verify/isValidAccessGrant.ts
```diff
@@ -48,7 +51,20 @@ async function isValidAccessGrant(
   } = {},
 ): Promise<{ checks: string[]; warnings: string[]; errors: string[] }> {
   const fetcher = await getSessionFetch(options);
-  const vcObject = await getBaseAccess(vc, options);
+  let validVC = null;
```
NSeydoux, line 54:
```suggestion
  let validVc = null;
```
That's minor, but introduces a naming inconsistency otherwise.

src/gConsent/guard/isDatasetCore.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @returns true if the candidate implements the DatasetCore interface, false otherwise.
+ * @internal
+ */
+function isDatasetCore(
+  candidateDataset: unknown | DatasetCore,
+): candidateDataset is DatasetCore {
+  return typeof (candidateDataset as DatasetCore).add === "function";
+}
+
+export { isDatasetCore };
```
NSeydoux, line 37:
```suggestion
export function isDatasetCore(
  candidateDataset: unknown | DatasetCore,
): candidateDataset is DatasetCore {
  return typeof (candidateDataset as DatasetCore).add === "function";
}
```
The export declarations are inlined in the rest of the codebase.

NSeydoux (changes requested):
Can you extract out the validation code into a shared `toVcDataset` that implements both the logic you have here (converting a JSON object into a `DatasetCore`) and lifts the dereferencing from `getBasedAccess`, i.e. factors out
```
if (typeof vc === "string" || vc instanceof URL) {
    baseVc = await getVerifiableCredential(vc.toString(), {
      returnLegacyJsonld: false,
      skipValidation: true,
      fetch: options.fetch,
    });
  }
``` 
? This way, `getBaseAccess` can be simplified to only take a `DatasetWithId` in, and `toVcDataset` is the single function that validates the input VC, either by fetching it or by converting it from plain JS object to `DatasetCore`. you can then reuse that `toVcDataset`
functions in each place you've implemented the validation logic in this PR.

Does that make sense?

NSeydoux:
Also, can you add a unit test to verify that converting from a plain JS object to a `DatasetCore` actually works?

src/common/verify/isValidAccessGrant.ts
garciafdezpatricia: Following the prior review, this change in the validations would be in `toVcDataset`?

NSeydoux: Absolutely

garciafdezpatricia: So, just to sum up, `vcToDataset` will either fetch the vc if it is a valid vc, that is:

`    if (typeof vc === "string" || vc instanceof URL) {
        return await getVerifiableCredential(vc.toString(), {
            returnLegacyJsonld: false,
            skipValidation: true,
            fetch: options.fetch,
        });
    }`

or, if it is not, convert it to a valid vc:

`if (
        typeof vc === "object" && 
        !isDatasetCore(vc) && 
        typeof (vc as { id: string }).id === "string"
    ) {
        return await verifiableCredentialToDataset(vc, {
            includeVcProperties: true,
            requireId: true,
        });
    }`

src/gConsent/manage/approveAccessRequest.ts
```diff
@@ -287,11 +289,17 @@ export async function approveAccessRequest(
...
+  if (validVc === undefined && requestVc && isUrl(requestVc.toString())) {
+    validVc = requestVc;
+  }
 
   const internalGrantOptions = initializeGrantParameters(
-    typeof requestVc !== "undefined"
+    typeof validVc !== "undefined"
       ? await getBaseAccess(
-          requestVc,
+          validVc as DatasetWithId,
```
NSeydoux, line 302:
```suggestion
          validVc,
```
ditto

src/gConsent/manage/denyAccessRequest.ts
```diff
@@ -91,8 +93,21 @@ async function denyAccessRequest(
     returnLegacyJsonld?: boolean;
   },
 ): Promise<DatasetWithId> {
+  let validVc;
+  validVc = await toVcDataset(vc, options);
```
NSeydoux, line 97:
```suggestion
  const validVc = await toVcDataset(vc, options);
```

src/gConsent/manage/denyAccessRequest.ts
```diff
@@ -91,8 +93,21 @@ async function denyAccessRequest(
...
+
+  if (validVc === undefined && isUrl(vc.toString())) {
+    validVc = vc;
+  }
```
NSeydoux, line 101:
```suggestion
```

src/gConsent/manage/denyAccessRequest.ts
```diff
@@ -91,8 +93,21 @@ async function denyAccessRequest(
...
+
+  if (validVc === undefined) {
+    throw new Error(
+      `Invalid argument: expected either a VC URL or a RDFJS DatasetCore, received ${vc}`,
+    );
+  }
+
   const baseVc: DatasetWithId = await getBaseAccess(
-    vc,
+    validVc as DatasetWithId,
```
NSeydoux, line 110:
```suggestion
    validVc,
```

src/gConsent/manage/revokeAccessGrant.ts
```diff
@@ -53,7 +55,24 @@ async function revokeAccessCredential(
   types: NamedNode<string>[],
   options: Omit<AccessBaseOptions, "accessEndpoint"> = {},
 ): Promise<void> {
-  const credential = await getBaseAccess(vc, options, types[0]);
+  let validVc;
+  validVc = await toVcDataset(vc, options);
```
NSeydoux, line 59:
```suggestion
  const validVc = await toVcDataset(vc, options);
```

src/gConsent/manage/revokeAccessGrant.ts
```diff
@@ -53,7 +55,24 @@ async function revokeAccessCredential(
...
+
+  if (validVc === undefined && isUrl(vc.toString())) {
+    validVc = vc;
+  }
```
NSeydoux, line 63:
```suggestion
```

src/gConsent/manage/revokeAccessGrant.ts
```diff
@@ -53,7 +55,24 @@ async function revokeAccessCredential(
...
+
+  if (validVc === undefined) {
+    throw new Error(
+      `Invalid argument: expected either a VC URL or a RDFJS DatasetCore, received ${vc}`,
+    );
+  }
+
+  const credential = await getBaseAccess(
+    validVc as DatasetWithId,
```
NSeydoux, line 72:
```suggestion
    validVc,
```

NSeydoux (changes requested):

src/common/verify/isValidAccessGrant.test.ts
```diff
@@ -112,6 +112,21 @@ describe("isValidAccessGrant", () => {
...
   });
 
+  it("uses the provided fetch if any from plain JSON VC", async () => {
+    jest.mocked(isVerifiableCredential).mockReturnValueOnce(true);
+    const mockedFetch = jest.fn<typeof fetch>().mockResolvedValueOnce(
+      new Response(JSON.stringify(MOCK_VERIFY_RESPONSE), {
+        status: 200,
+      }),
+    );
+    await isValidAccessGrant(JSON.parse(JSON.stringify(MOCK_ACCESS_GRANT)), {
```
NSeydoux, line 122:
```suggestion
    await isValidAccessGrant(MOCK_ACCESS_GRANT_BASE), {
```
Although what you had isn't incorrect, since `MOCK_ACCESS_GRANT` is actually derived from `MOCK_ACCESS_GRANT_BASE`, so we might as well use it directly.

src/common/verify/isValidAccessGrant.test.ts
```diff
@@ -112,6 +112,21 @@ describe("isValidAccessGrant", () => {
     expect(mockedFetch).toHaveBeenCalled();
   });
 
+  it("uses the provided fetch if any from plain JSON VC", async () => {
```
NSeydoux, line 115: Unfortunately, this test doesn't really assert anything on the value passed to the verify endpoint, so it's not a great candidate to test for the plain JSON VC. That's on me, there wasn't a good way for you to know that in advance.

I think you can remove this test altogether, as the other one in this file covers the same use-case, but in a more meaningful way.

src/common/verify/isValidAccessGrant.test.ts
```diff
@@ -137,6 +152,31 @@ describe("isValidAccessGrant", () => {
...
+      specCompliant: {},
+      legacy: {},
+    });
+    const mockedFetch = jest.fn<typeof fetch>().mockResolvedValueOnce(
+      new Response(JSON.stringify(MOCK_VERIFY_RESPONSE), {
+        status: 200,
+      }),
+    );
+
+    await isValidAccessGrant(MOCK_ACCESS_GRANT, {
```
NSeydoux, line 168:
```suggestion
    await isValidAccessGrant(MOCK_ACCESS_GRANT_BASE, {
```

src/common/verify/isValidAccessGrant.test.ts
```diff
@@ -184,6 +248,32 @@ describe("isValidAccessGrant", () => {
...
+    await isValidAccessGrant(
+      JSON.parse(JSON.stringify("https://example.com/someVc")),
+      {
+        fetch: mockedFetch,
+        verificationEndpoint: MOCK_ACCESS_ENDPOINT,
+      },
+    );
+
+    expect(mockedFetch).toHaveBeenCalledWith("https://example.com/someVc");
+  });
```
NSeydoux, line 275:
```suggestion
```
In this test, it's not an Access Grant that is passed, but a URL, so we are already fetching and parsing the VC as we see fit. There is nothing to add to the original test.

src/common/verify/isValidAccessGrant.test.ts
```diff
@@ -202,6 +292,27 @@ describe("isValidAccessGrant", () => {
...
+        JSON.parse(JSON.stringify("https://example.com/someVc")),
+        {
+          fetch: mockedFetch,
+          verificationEndpoint: MOCK_ACCESS_ENDPOINT,
+        },
+      ),
+    ).rejects.toThrow(
+      "Verifiable credential is not an object, or does not have an id",
+    );
+  });
```
NSeydoux, line 314:
```suggestion
```
Same as before, this is for a URL to dereference a VC, not a VC payload.

src/gConsent/manage/approveAccessRequest.ts
```diff
@@ -22,10 +22,10 @@
...
 import { acp_ess_2 } from "@inrupt/solid-client";
-import type {
-  DatasetWithId,
-  VerifiableCredential,
-  VerifiableCredentialBase,
+import {
+  type DatasetWithId,
+  type VerifiableCredential,
+  type VerifiableCredentialBase,
 } from "@inrupt/solid-client-vc";
```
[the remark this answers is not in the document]
garciafdezpatricia: I'd say that would be the automatic imports, sorry about that

src/common/verify/isValidAccessGrant.test.ts
```diff
@@ -227,6 +338,31 @@ describe("isValidAccessGrant", () => {
...
+      vcModule,
+      "getVerifiableCredentialApiConfiguration",
+    );
+    jest.mocked(isVerifiableCredential).mockReturnValueOnce(true);
+    const mockedFetch = jest.fn<typeof fetch>().mockResolvedValueOnce(
+      new Response(JSON.stringify(MOCK_VERIFY_RESPONSE), {
+        status: 200,
+      }),
+    );
+    await isValidAccessGrant(JSON.parse(JSON.stringify(MOCK_ACCESS_GRANT)), {
```
NSeydoux, line 355:
```suggestion
    await isValidAccessGrant(MOCK_ACCESS_GRANT_BASE), {
```

src/common/verify/isValidAccessGrant.test.ts
```diff
@@ -256,6 +392,35 @@ describe("isValidAccessGrant", () => {
...
+        new Response(JSON.stringify(MOCK_ACCESS_GRANT), { status: 200 }),
+      )
+      // Then, the verification endpoint is called
+      .mockResolvedValueOnce(
+        new Response(JSON.stringify(MOCK_VERIFY_RESPONSE), {
+          status: 200,
+        }),
+      );
+
+    await isValidAccessGrant(JSON.parse(JSON.stringify(MOCK_ACCESS_GRANT)), {
```
NSeydoux, line 417:
```suggestion
    await isValidAccessGrant(MOCK_ACCESS_GRANT_BASE), {
```

src/common/verify/isValidAccessGrant.test.ts
garciafdezpatricia: I tried with that, but since `isValidAccessGrant` input allows types are either `string`, `VerifiableCredential`, `DatasetWithId` or `URL`, with `MOCK_ACCESS_GRANT_BASE` being just an object, it raises a TS error.

src/gConsent/manage/revokeAccessGrant.test.ts
```diff
@@ -112,6 +136,20 @@ describe("revokeAccessGrant", () => {
...
+        status: 401,
+        statusText: "Unauthorized",
+      }),
+    );
+    await expect(
+      revokeAccessGrant(JSON.parse(JSON.stringify("https://some.credential")), {
+        fetch: mockedFetch,
+      }),
+    ).rejects.toThrow(/\[https:\/\/some.credential\].*401.*Unauthorized/);
+  });
```
NSeydoux, line 151:
```suggestion
```
Ditto

src/gConsent/manage/revokeAccessGrant.test.ts
```diff
@@ -142,4 +190,29 @@ describe("revokeAccessGrant", () => {
...
+        fetch: mockedFetch,
+      },
+    );
+    expect(spiedRevoke).toHaveBeenCalledWith(
+      "https://some.issuer/status",
+      MOCKED_CREDENTIAL_ID,
+      expect.anything(),
+    );
+    expect(mockedFetch).not.toHaveBeenCalled();
+  });
```
NSeydoux, line 217:
```suggestion
```
Ditto

src/common/verify/isValidAccessGrant.test.ts
NSeydoux: Oh I see, then stringify and parse is fine, or you can do `as any`, which is equivalent. ESlint will complain about it, you can add an ignore mentioning we are explicitly testing for a behavior that is not compliant with our API type definitions.

src/gConsent/discover/redirectToAccessManagementUi.ts
```diff
@@ -20,9 +20,9 @@
 //
 
 import type { UrlString, WebId } from "@inrupt/solid-client";
-import type {
-  DatasetWithId,
-  VerifiableCredential,
+import {
+  type DatasetWithId,
+  type VerifiableCredential,
```
NSeydoux, line 25: Is this change intentional, or is it introduced by `prettier`?

NSeydoux (approved):

src/gConsent/discover/redirectToAccessManagementUi.ts
garciafdezpatricia: It wasn't intentional at all. I believe it has been introduced by `prettier` indeed

PR #1077, opened 2024-08-12
Ignore major version updates to typedoc-plugin-markdown and eslint

acoburn:
The major version upgrades to the `eslint` and `typedoc-plugin-markdown` dependencies require some more significant refactors of the existing code. For now, these updates will be turned off to produce less dependabot noise.


edwardsph (approved):
Have you checked whether these dependencies are also used in the test app package?

acoburn:
@edwardsph they are not listed in any of the package.json files in `examples/` or in `e2e/`

PR #1078, opened 2024-08-16
SDK-3357: Add node 22 to matrix and set as default

edwardsph:
# New feature description

Added Node 22 support.


.github/workflows/e2e-browser.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 20.x
+          node-version: 22.x
```
NSeydoux, line 29: Here, the node version is only ever used to build, it's irrelevant to the tests. I suggest we change this to `latest`, to avoid the churn.

NSeydoux (approved):

.github/workflows/e2e-browser.yml
edwardsph: See #1079

PR #1100, opened 2024-08-26
Align wrapper signatures with underlying function

NSeydoux:
This prevents misalignment between the `/resource` module functions signatures and the underlying `@inrupt/solid-client` functions they wrap around. The `/resource` module directly reuses the signatures of the wrapped function instead of redefining its own, which updates the signature on dependency upgrade.

The resulting API docs is not ideal, but this is an issue that will be investigated separately.


NSeydoux:
I hit send a bit too fast on [5c7c28e](https://github.com/inrupt/solid-client-access-grants-js/pull/1100/commits/5c7c28ef5ce652199abedbb07ab4500af452e41c), this actually fixes unit tests as well.

edwardsph (approved):
LGTM

PR #1190, opened 2024-12-13
Throw custom error

NSeydoux:
This makes it so that the library throws a custom `AccessGrantError` (or `UmaError`, extending the shared `InruptClientError` instead of a generic `Error`.

### DIFF
--- a/src/gConsent/util/issueAccessVc.ts
+++ b/src/gConsent/util/issueAccessVc.ts
@@ -53,6 +53,7 @@ import { accessToResourceAccessModeArray } from "./accessToResourceAccessModeArr
 import { isBaseRequest } from "../guard/isBaseRequest";
 import type { AccessCredentialType } from "../type/AccessCredentialType";
 import type { CustomField } from "../../type/CustomField";
+import { AccessGrantError } from "../../common/errors/AccessGrantError";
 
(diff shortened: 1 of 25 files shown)


src/gConsent/util/issueAccessVc.ts
```diff
@@ -210,13 +211,13 @@ export async function issueAccessVc(
     hasConsent &&
     vcBody.credentialSubject.hasConsent.forPersonalData.length <= 0
   ) {
-    throw new Error("There are no resources in the access request!");
+    throw new AccessGrantError("There are no resources in the access request!");
```
edwardsph, line 214:
```suggestion
    throw new AccessGrantError("There are no resources in the access request");
```
For consistency with the related message but then there is also a question of consistency throughout. Should error messages end with punctuation or not?

edwardsph (approved):
LGTM

src/gConsent/util/issueAccessVc.ts
NSeydoux: I don't have a strong opinion on this. I would say they should not, because some of them will end wrapping another error message sent by an underlying library, which we do not control and cannot know whether it ends with punctuation or not.

PR #1195, opened 2024-12-17
Access grant queries

NSeydoux:


package.json
```diff
@@ -131,7 +135,7 @@
   "dependencies": {
     "@inrupt/solid-client": "^2.0.0",
     "@inrupt/solid-client-errors": "^0.0.2",
-    "@inrupt/solid-client-vc": "^1.1.2",
+    "@inrupt/solid-client-vc": "^1.2.0",
```
NSeydoux, line 138: This adds query endpoint discovery, but its usage is currently blocked by a small ESS bug.

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+ * const activeGrantsToday2 = await query(
+ *       activeGrantsToday.next,
+ *       {
+ *         fetch: session.fetch,
+ *         queryEndpoint: config.queryEndpoint,
+ *       },
+ *     );
+ * ```
+ */
+export async function query(
```
NSeydoux, line 245: This is the main addition to the public API. The rest of this file is not exported.

CHANGELOG.md
```diff
@@ -6,18 +6,26 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 The following changes are pending, and will be applied on the next major release:
 
-- The `status` parameter for `getAccessGrantAll` will default to `all` rather than `granted`.
+- The `getAccessGrantAll` function is deprecated. It can be replaced with `query`.
+  Although the two functions behavior is different, they can be used to achieve
+  similar results.
+- The `gConsent` and all of `gConsent/*` submodules are deprecated. The former can
+  be replaced by a regular import of the library, and the latters can be replaced
```
edwardsph, line 13:
```suggestion
  be replaced by a regular import of the library, and the latter can be replaced
```
Weird English but latter is also plural

CHANGELOG.md
```diff
@@ -6,18 +6,26 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
...
+### New feature
 
 - Add support for custom fields. Applications are now able to read and write custom fields
   into Access Credentials (both Access Requests and Access Grants). This feature is available
   via a new option introduced in `issueAccessRequest` and `approveAccessRequest` to write the
   custom fields, and via a set of dedicated getters in the `getters/` module. A generic getter
   is introduced, `getCustomFields`, as well as a set of typed helpers, such as `getCustomInteger`.
   Typed helpers are available for integers, floats, strings and booleans.
+- Support new query endpoint: the new `query` function allows to query for Access Credentials using
+  the newly introduce ESS endpoint.
```
edwardsph, line 28:
```suggestion
- Support new query endpoint: the new `query` function enables querying for Access Credentials using
  the newly introduced ESS endpoint.
```

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+
+import LinkHeader from "http-link-header";
+
+import { handleErrorResponse } from "@inrupt/solid-client-errors";
+import type { DatasetWithId } from "@inrupt/solid-client-vc";
+import { verifiableCredentialToDataset } from "@inrupt/solid-client-vc";
+import type { UrlString } from "@inrupt/solid-client";
+
+/**
+ * Supported Access Credentials status
```
edwardsph, line 30:
```suggestion
 * Supported Access Credential statuses
```

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+  /**
+   * URL of the resource featured in the Access Credential.
+   */
+  resource?: UrlString;
+  /**
+   * URL of the Access Credential purpose.
+   */
+  purpose?: UrlString;
+  /**
+   * Interval (expressed using ISO 8601) during which the Credential was issued.
```
edwardsph, line 85:
```suggestion
   * Period (expressed using ISO 8601) during which the Credential was issued.
```
I think this is slightly better - this post suggests the differences: https://www.quora.com/What-is-the-difference-between-a-period-duration-and-interval

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+   */
+  issuedWithin?: "P1D" | "P7D" | "P1M" | "P3M";
+  /**
+   * Interval (expressed using ISO 8601) during which the Credential was revoked.
```
edwardsph, line 89:
```suggestion
   * Period (expressed using ISO 8601) during which the Credential was revoked.
```

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+export const DURATION = {
+  ONE_DAY: "P1D",
+  ONE_WEEK: "P7D",
+  ONE_MONTH: "P1M",
+  THREE_MONTHS: "P3M",
+} as const;
+
+export type CredentialFilter = {
+  /**
+   * The Access Credential type (e.g. Access Request, Access Grant...)
```
edwardsph, line 61:
```suggestion
   * The Access Credential type (e.g. Access Request, Access Grant...).
```

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+   */
+  type?: CredentialType;
+  /**
+   * The Access Credential status (e.g. Active, Revoked...)
```
edwardsph, line 65:
```suggestion
   * The Access Credential status (e.g. Active, Revoked...).
```

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+  const filter: CredentialFilter = {};
+  url.searchParams.forEach((value, key) => {
+    if (isSupportedFilterElement(key)) {
+      Object.assign(filter, { [`${key}`]: value });
+    }
+  });
+  return filter;
+}
+
+function hasItems(x: unknown): x is { items: unknown[] } {
```
edwardsph, line 159: This appears to test that items exists and is an array, so `hasItems` means it has the items field correctly defined as opposed to there are items present in the result. Is that confusing? I initially thought it was going to include `items.length > 0` in the result.

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+  const candidate = x as { items: unknown };
+  return candidate.items !== undefined && Array.isArray(candidate.items);
+}
+
+function isObjectWithId(x: unknown): x is { id: string } {
+  const candidate = x as { id: string };
+  return typeof candidate.id === "string";
```
edwardsph, line 166: Does it need to test that the `id` is a non-empty string?

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+      if (link.length === 0) {
+        return;
+      }
+      result[rel] = toCredentialFilter(new URL(link[0].uri));
+    });
+  }
+  const body = await response.json();
+  if (!hasItems(body) || !body.items.every(isObjectWithId)) {
+    throw Error(
+      `Unexpected response, no items found in ${JSON.stringify(body)}`,
```
edwardsph, line 192: Based on my previous comment, this would throw if `items` as not defined as an error but would not throw if it was empty although the error message looks like it would cover that condition.

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+  Object.entries(filter).forEach(([key, value]) => {
+    if (isSupportedFilterElement(key)) {
+      result.searchParams.append(key, value.toString());
+    }
+  });
+  return result;
+}
+
+/**
+ * Query for Access Requests or Access Grants based on a given filter.
```
edwardsph, line 214: It's a little weird that we talk about Access Requests or Access Grants everywhere but never mention Denials which are also included in the query function. This is a much wider observation than this PR.

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+ *
+ * @param filter the query filter
+ * @param options query options
```
edwardsph, line 217:
```suggestion
 * @param filter The query filter
 * @param options Query options
```
We mostly seem to capitalize this but are not consistent about punctuation but I on't know if we actually have a standard.

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+ * @returns a paginated set of Access Credential matching the given filter
+ * @since unreleased
+ *
+ * @example
+ * ```
+ *  // Get the first results page.
+ *  const activeGrantsToday = await query(
```
edwardsph, line 224:
```suggestion
 *  const activeGrantsWithinDay = await query(
```
It is not related to a time so this might be clearer

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+ *         status: "Active",
+ *         issuedWithin: DURATION.ONE_DAY,
+ *       },
+ *       {
+ *         fetch: session.fetch,
+ *         queryEndpoint: config.queryEndpoint,
+ *       },
+ *     );
+ * // Get the next results page.
+ * const activeGrantsToday2 = await query(
```
edwardsph, line 236:
```suggestion
 * const activeGrantsWithinDay2 = await query(
```

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+ *       activeGrantsToday.next,
```
edwardsph, line 237:
```suggestion
 *       activeGrantsWithinDay.next,
```

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,262 @@
...
+): Promise<CredentialResult> {
+  const result: CredentialResult = { items: [] };
+  const linkHeader = response.headers.get("Link");
+  if (linkHeader !== null) {
+    const parsedLinks = LinkHeader.parse(linkHeader);
+    PAGING_RELS.forEach((rel) => {
+      const link = parsedLinks.get("rel", rel);
+      if (link.length > 1) {
+        throw Error(
+          `Unexpected response, found more than one ${rel} Link headers.`,
```
edwardsph, line 180:
```suggestion
          `Unexpected response, found more than one [${rel}] Link headers.`,
```
Makes it easier to read, as you did with the query endpoint error

src/gConsent/query/query.ts
NSeydoux: Just to make sure I understand the confusion correctly, what creates confusion is that the name of the function suggests we are checking for a non-zero array, when it is only a type guard? I don't mind renaming it to anything else, it's just an internal function anyway. The complete signature showing that it i a type guard, I think it may limit the confusion, but I see your point.

I've seen two conventions for type guards: `hasX` when checking for a single attribute (like here) and `isX` when checking for multiple attributes making up a type. This could be renamed `isQueryResult` for instance, to remain idiomatic TS.

src/gConsent/query/query.ts
NSeydoux: That is a mistake from me in the message phrasing, an empty array is acceptable.

src/gConsent/query/query.ts
NSeydoux: Good point, I'll change that.

edwardsph (commented):
Looks good, just some non-functional comments

src/gConsent/query/query.ts
edwardsph: You've understood my concern and it comes from someone not used to type guards, so I'm OK if you think it is not an issue. I thought you might respond as you did which suggests that with a little more experience I would have had no question about this.

src/gConsent/query/query.ts
edwardsph: Again - this is a type guard, not a validator of the data so I think you can ignore my comment

NSeydoux: Well it depends on the constraints we put on the target type really, in this case I think that's a good point. I was using it as you describe (for validating data structure only), but here it might be more correct to check the validity of the `id` (specifically, that it is a URL)

src/gConsent/query/query.ts
NSeydoux: Refactoring this in light of this comment, I actually ended up removing the type guards: 87c5e87e8d16028db4f11d0dc54ae488704ad08b

src/gConsent/query/query.ts
NSeydoux: This function was incidentally removed as part of 87c5e87e8d16028db4f11d0dc54ae488704ad08b

src/gConsent/query/query.ts
NSeydoux: Changed in 87c5e87e8d16028db4f11d0dc54ae488704ad08b

src/gConsent/query/query.ts
NSeydoux: 7d323f9f0751b3a8b4495891ffbbd689fb0915d6

CHANGELOG.md
```diff
@@ -6,18 +6,26 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 The following changes are pending, and will be applied on the next major release:
 
-- The `status` parameter for `getAccessGrantAll` will default to `all` rather than `granted`.
+- The `getAccessGrantAll` function is deprecated. It can be replaced with `query`.
+  Although the two functions behavior is different, they can be used to achieve
```
edwardsph, line 10:
```suggestion
  Although the two functions' behavior is different, they can be used to achieve
```

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,267 @@
...
+      Object.assign(filter, { [`${key}`]: value });
+    }
+  });
+  return filter;
+}
+
+async function parseQueryResponse(
+  responseJson: unknown,
+): Promise<DatasetWithId[]> {
+  // The type assertion here is immediatly checked.
```
edwardsph, line 163:
```suggestion
  // The type assertion here is immediately checked.
```

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,267 @@
...
+  Object.entries(filter).forEach(([key, value]) => {
+    if (isSupportedFilterElement(key)) {
+      result.searchParams.append(key, value.toString());
+    }
+  });
+  return result;
+}
+
+/**
+ * Query for Access Credential (Access Requests, Access Grants or Access Denials) based on a given filter.
```
edwardsph, line 219:
```suggestion
 * Query for Access Credentials (Access Requests, Access Grants or Access Denials) based on a given filter.
```

src/gConsent/query/query.ts
```diff
@@ -0,0 +1,267 @@
...
+ *
+ * @param filter The query filter
+ * @param options Query options
+ * @returns a paginated set of  matching the given filter
```
edwardsph, line 223:
```suggestion
 * @returns a paginated set of Access Credentials matching the given filter
```

edwardsph (commented):
Last comments

edwardsph (approved):
LGTM

PR #1198, opened 2024-12-18
 Add async iterator for query results 

NSeydoux:
This adds a helper function to make it easier to go through all of the paginated results. It allows syntax like the following:
```js
  const pages = paginateQuery({ type: "SolidAccessGrant", status: "Active" },
    {
      fetch: session.fetch,
      queryEndpoint: new URL("https://vc.inrupt.com/query"),
    });
  let i = 0;
  for await (const page of pages) {
    console.log(`Page ${i} has ${page.items.length} items.`)
    i += 1;
  }
```

Unit tests aren't completely thorough, because technically they should check for all the error conditions, which are already checked in the underlying `query` function. I don't think it's worth the overhead.

### DIFF
--- a/src/gConsent/query/query.ts
+++ b/src/gConsent/query/query.ts
@@ -282,5 +282,5 @@ export async function* paginateQuery(
     page = await query(page.next, options);
   }
   // Return the last page.
-  yield page;
+  return page;
 }
--- a/src/gConsent/query/query.ts
+++ b/src/gConsent/query/query.ts
@@ -216,7 +216,8 @@ function toQueryUrl(endpoint: URL, filter: CredentialFilter): URL {
 }
 
 /**
- * Query for Access Credentials (Access Requests, Access Grants or Access Denials) based on a given filter.
+ * Query for Access Credential (Access Requests, Access Grants or Access Denials) based on a given filter,
+ * and get a page of results.
  *
  * @param filter The query filter
  * @param options Query options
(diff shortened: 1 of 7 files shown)


edwardsph (commented):
Really nice approach. Ideally we'd have tests that covered 0, 1, and 2 pages to account for different conditions in the generator

jholleran (approved):

src/gConsent/query/query.ts
```diff
@@ -265,3 +266,45 @@ export async function query(
...
+  let page = await query(filter, options);
+  while (page.next !== undefined) {
+    yield page;
+    // This is a generator, so we don't want to go through
+    // all the pages at once with a Promise.all approach.
+    // eslint-disable-next-line no-await-in-loop
+    page = await query(page.next, options);
+  }
+  // Return the last page.
+  yield page;
```
edwardsph, line 309: You previously had return here and I looked up why. It not only yields the last page but marks the iterator as done so I thought you were right. Why do you think it should be yield?

NSeydoux: That's what I initially expected too, and that's why I had a return, but in tests that didn't work as I expected. If I understood correctly, the iterator has a `return` function to mark it done, but here the function is a generator, which returns said iterator, hence the difference. None of the generator examples in the MDN docs use return, and here is an example in the Node CLI:
```
> function* test() {
... yield 1;
... yield 2;
... yield 3;
... return 4;
... }
undefined
> const i = test()
undefined
> Array.from(i)
[ 1, 2, 3 ]
```

PR #1201, opened 2024-12-22
Add restrictions on Access Crendentials queries

NSeydoux:
Queries must now be on Access Grants or Access Requests, rather than generically on Acces Credentials. This is aligned with the JCL general behavior.

This also enforces URL objects are used rather than string, which is more consistent with the rest of the API (in particular, URL are used in custom fields already.


src/gConsent/query/query.ts
```diff
@@ -24,20 +24,22 @@ import LinkHeader from "http-link-header";
 import { handleErrorResponse } from "@inrupt/solid-client-errors";
 import type { DatasetWithId } from "@inrupt/solid-client-vc";
 import { verifiableCredentialToDataset } from "@inrupt/solid-client-vc";
-import type { UrlString } from "@inrupt/solid-client";
 import { AccessGrantError } from "../../common/errors/AccessGrantError";
 
 /**
- * Supported Access Credential statuses
+ * Supported Access Requests statuses
```
edwardsph, line 30:
```suggestion
 * Supported Access Request statuses
```

edwardsph (approved):
LGTM

PR #1209, opened 2025-01-12
Use the v2 Access Grant JSON-LD context

NSeydoux:


src/gConsent/constants.ts
NSeydoux, line 1: This is the core change of the PR, where the canonical context is updated and the self-hosted removed.

src/gConsent/manage/getAccessGrantAll.test.ts
NSeydoux, line 1: There were no test assertions on the context used when calling the derive endpoint. This adds these, in addition of a light refactoring.

src/gConsent/request/issueAccessRequest.test.ts
```diff
@@ -244,54 +239,6 @@ describe.each([true, false, undefined])(
       ).rejects.toThrow();
     });
 
-    it("computes the correct context based on the issuer", async () => {
```
NSeydoux, line 247: These tests checked for the self-hosted context, which is no longer included.

.github/workflows/e2e-node.yml
```diff
@@ -19,7 +19,8 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         node-version: ["18.x", "20.x", "22.x"]
-        environment-name: ["ESS PodSpaces", "ESS Dev-2-2"]
+        # FIXME: re-enable ESS 2.2 tests
+        environment-name: ["ESS PodSpaces"]
```
NSeydoux, line 23: Temporary workaround to release the fix quickly, and then re-enabled the tests.

NSeydoux:
Do not merge before backwards compatibility is added (using the `.well-known/vc-configuration`)

src/gConsent/manage/approveAccessRequest.test.ts
```diff
@@ -145,6 +146,9 @@ describe("approveAccessRequest", () => {
     accessGrantVc = await mockAccessGrantVc();
     consentRequestVc = await mockConsentRequestVc();
     consentGrantVc = await mockConsentGrantVc();
+    cacheProvider(new URL(MOCKED_ACCESS_ISSUER).href, {
```
NSeydoux, line 149: There is a bunch of changes like this in this PR, it pre-fills the cache so that the test doesn't go to fetch the mock VC provider config.

NSeydoux (commented):
There are a couple of FIXMEs in this PR that are not urgent, they can be cleaned up after this has been released.

src/fetch/index.ts
```diff
@@ -121,6 +119,7 @@ export async function exchangeTicketForAccessToken(
 
   try {
     const data = await response.json();
+    console.log(data);
```
edwardsph, line 122:
```suggestion
```
I assume this was here while testing

edwardsph (approved):
LGTM

PR #1259, opened 2025-04-07
SDK-3395: Link Access Grant to Access Request

NSeydoux:
When issuing an Access Grant from an Access Request, there is now a formal link from the former to the latter now, which results in an improved behavior of the query endpoint.

### DIFF
--- a/e2e/node/e2e.test.ts
+++ b/e2e/node/e2e.test.ts
@@ -88,6 +88,8 @@ import {
   getPurposes,
 } from "../../src/common/getters";
 import { toBeEqual } from "../../src/gConsent/util/toBeEqual.mock";
+import { setGlobalDispatcher, ProxyAgent } from "undici";
+
 
 const { namedNode } = DataFactory;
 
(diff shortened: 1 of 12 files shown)


e2e/node/e2e.test.ts
```diff
@@ -1774,29 +1774,76 @@ describe(`End-to-end access grant tests for environment [${environment}] `, () =
...
+            expirationDate: new Date(Date.now() + 60 * 60 * 1000),
+          },
+          {
+            fetch: addUserAgent(requestorSession.fetch, TEST_USER_AGENT),
+            accessEndpoint: vcProvider,
+            updateAcr: false,
+            returnLegacyJsonld: false,
+          },
+        );
+        console.debug("Access request issued");
```
jholleran, line 1811: Do you need this debug log?

jholleran (approved):

e2e/node/e2e.test.ts
NSeydoux: It was forgotten when cleaning up the tests debug, removed in c573ab0cca099201238f5effb04236cc35478892.

PR #1275, opened 2025-04-28
Fix incomplete e2e test

NSeydoux:
The test checking that requests are no longer pending after a grant is issued had a bug making it not properly checking that the status of the request was appropriately applied after the grant has been issued. The feature was working and is unchanged, this is just a fix to the test so that it is correct.


edwardsph (approved):
LGTM

